Issue #186, opened 2020-01-13
Reduce image size of README 


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- Current images are uploaded in 
https://github.com/DroidKaigi/conference-app-2020/issues/74#issuecomment-573555849

- Those images are bigger than specified by the width of the `img` tag.
- I want to reduce the size of it

 ## Links	

-

MasayukiSuda:
🙋‍♂

takahirom:
Assigned🙏🙏🙏

FujiKinaga:
Fixed by #221 💯

PR #221, opened 2020-01-14
Replace Readme image with reduced image

MasayukiSuda:
## Issue
- https://github.com/DroidKaigi/conference-app-2020/issues/186

## Overview (Required)
- Replace Readme image with reduced image
  - reduced the size by about 60% while maintaining the quality.


takahirom:
Thank you for lightweight images! We can read the README on Android😆

PR #230, opened 2020-01-14
Fixed a fragment view leak by using DataBinding-ktx

### LINKED ISSUE #153: Fix fragment view leaks
## Kind (Required)

- Improvement

 ## Overview (Required)

- When the Fragment is on the stack, it seems to be leaking because the DataBinding field of the Fragment exists 🦅 
 - It's best to do this after fixing the issue in #149.

TomoyaShibata:
## Issue
- close #153

## Overview (Required)

- References to view remain unintentionally when add a fragment to a backstack.
- You can destroy the reference by setting `binding = null` in `Fragment#onDestroyView` lifecycle, but introduced [wada811/DataBinding-ktx](https://github.com/wada811/DataBinding-ktx) which does it automatically.

## Links

- https://github.com/wada811/DataBinding-ktx
- https://twitter.com/Piwai/status/1201686009349455872

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ allprojects {
         maven { url 'https://oss.sonatype.org/content/repositories/snapshots/' }
         maven { url "https://dl.bintray.com/kotlin/ktor" }
         maven { url "https://dl.bintray.com/kotlin/kotlinx" }
+        maven { url 'https://jitpack.io' }
     }
     plugins.whenPluginAdded {
         if (it.isAndroidApp() || it.isAndroidLibrary() || it.isDynamicFeature()) {
--- a/buildSrc/src/main/java/dependencies/Dep.kt
+++ b/buildSrc/src/main/java/dependencies/Dep.kt
@@ -222,4 +222,9 @@ object Dep {
     object Insetter {
         val insetter = "dev.chrisbanes:insetter-ktx:0.2.0"
     }
+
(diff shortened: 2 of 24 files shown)


takahirom:
👀

takahirom:
DataBinding-ktx is awesome🔥
But if we want to use like recyclerview adapter. We should assign null
How about this?
https://github.com/android/architecture-components-samples/blob/9826b59956eb93c9e627bdf16a19a1c8bc28ce14/GithubBrowserSample/app/src/main/java/com/android/example/github/util/AutoClearedValue.kt#L31-L40
I think it is OK to use DataBinding-ktx , but I am thinking.🤔

TomoyaShibata:
AutoClearedValue awsome idea too! I will try it's approach.

Another great feature of DataBinding-ktx is the automatic invocation of setLifecycleOwner, but this is not required for this app. So it is possible to unify to AutoClearedValue.

TomoyaShibata:
Add what you care about. :memo:

There are comments in the source code that can create situations that cause NPE to be thrown, and This tweet is  have mentioned it. :eyes:

https://twitter.com/wada811/status/1217361894224220160

takahirom:
Thank you for investigate :pray: 
If you have time, can you create AutoClearedValue version PR?
I don't think it's a big problem. 🙇

PR #243, opened 2020-01-14
Add navigation transitions

### LINKED ISSUE #196: Transition with fade
## Kind (Required)

- Improvement

 ## Overview (Required)

- Android dev summit app seems to be faded at the time of screen transition. Can we do the same?

 ## Links	

- https://play.google.com/store/apps/details?id=com.google.samples.apps.adssched&hl=ja

rmakiyama:
## Issue
- close #196

## Overview (Required)
- add fade in/out transition.
- `config_navDefaultAnimTime` referenced `config_activityDefaultDur` in sdk.

I can not good work for `toolbar` and `MainSessionsFragment`.
Toolbar color does not change color smooth 🙇 
And, I want to add `popEnterAnim` to MainSessionsFragment actions but It is not good working...
I want to create these another issues.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


corecomponent/androidcomponent/src/main/res/values/integers.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <integer name="config_navDefaultAnimTime">220</integer>
```
takahirom: Can you add this comment? (It is a little sudden) 💬 

```
<! -- referenced config_activityDefaultDur in sdk -->
```

corecomponent/androidcomponent/src/main/res/navigation/navigation.xml
```diff
@@ -11,14 +11,23 @@
         <action
             android:id="@+id/action_session_to_session_detail"
             app:destination="@id/session_detail"
+            app:enterAnim="@anim/fade_in"
```
takahirom: Can we set default animation? 👀

corecomponent/androidcomponent/src/main/res/values/integers.xml
rmakiyama: yes, thanks your advice! 🙏

corecomponent/androidcomponent/src/main/res/navigation/navigation.xml
rmakiyama: I tried use override default file like [this](https://github.com/android/architecture-components-samples/issues/537) but It not working...
Do you have any idea?

takahirom: Can you please add an issue that "defines the default animation"?

rmakiyama: I add an issue🙏🏻 #245

PR #246, opened 2020-01-14
Session search/filter box and result section headers are now localizable

atsushieno:
The query box was Japanese, and the result section headers were English,
both in direct strings.


takahirom:
Great! Thanks!🙏

PR #252, opened 2020-01-15
When landscape, exclude navigation bar area

### LINKED ISSUE #203: When landscape, we can not show the navigation drawer
## Kind (Required)

- Improvement

 ## Overview (Required)

- We should apply window inset
image

 ## Links	

-

fornewid:
## Issue
- close #203

## Overview (Required)
- Set paddings to contents container using insetter.
- On seascape mode only, nav bar is overlapped with DrawerLayout.
   So set left padding and reset width for NavigationView.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
+++ b/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
@@ -3,11 +3,13 @@ package io.github.droidkaigi.confsched2020
 import android.annotation.SuppressLint
 import android.os.Bundle
 import android.view.View
+import android.view.ViewGroup
 import androidx.annotation.IdRes
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.updateLayoutParams
+import androidx.core.view.updatePadding
 import androidx.databinding.DataBindingUtil
 import androidx.drawerlayout.widget.DrawerLayout
 import androidx.fragment.app.FragmentActivity
@@ -92,8 +94,21 @@ class MainActivity : DaggerAppCompatActivity() {
         setupNavigation()
         setupStatusBarColors()
 
-        binding.drawerLayout.doOnApplyWindowInsets { _, insets, _ ->
+        binding.drawerLayout.doOnApplyWindowInsets { _, insets, initialState ->
             binding.drawerLayout.setChildInsetsWorkAround(insets)
+            binding.contentContainer.updatePadding(
+                left = insets.systemWindowInsetLeft + initialState.margins.left,
+                right = insets.systemWindowInsetRight + initialState.margins.right
+            )
+            binding.navView.apply {
+                // On seascape mode only, nav bar is overlapped with DrawerLayout.
+                // So set left padding and reset width.
+                val leftSpace = insets.systemWindowInsetLeft + initialState.margins.left
+                updatePadding(left = leftSpace)
+                updateLayoutParams<ViewGroup.MarginLayoutParams> {
+                    width = resources.getDimensionPixelSize(R.dimen.nav_drawer_width) + leftSpace
+                }
+            }
         }
         binding.toolbar.doOnApplyWindowInsets { _, insets, initialState ->
             binding.toolbar.updateLayoutParams<ConstraintLayout.LayoutParams> {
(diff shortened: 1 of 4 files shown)


yurihondo:
👀

android-base/src/main/java/io/github/droidkaigi/confsched2020/ui/widget/SystemUiManager.kt
```diff
@@ -58,8 +58,11 @@ class SystemUiManager(
             0
         } else {
             View.SYSTEM_UI_FLAG_LIGHT_STATUS_BAR
+        } or if (isLandscape()) {
```
yurihondo: @fornewid 
Thank you for your PR.

Changing the flags will affect the entire application, so I would like to achieve the same behavior using the insetter if possible, like so
https://github.com/DroidKaigi/conference-app-2020/pull/72/files#diff-d0a86a9871834b5287736b83beab9d37R79
Could you consider this?

fornewid: Thanks, I'll reconsider using insetter ASAP. :)

yurihondo: Take as much time as you need :D
Thank you for your cooperation!

android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
```diff
@@ -92,8 +94,21 @@ class MainActivity : DaggerAppCompatActivity() {
         setupNavigation()
         setupStatusBarColors()
 
-        binding.drawerLayout.doOnApplyWindowInsets { _, insets, _ ->
+        binding.drawerLayout.doOnApplyWindowInsets { _, insets, initialState ->
             binding.drawerLayout.setChildInsetsWorkAround(insets)
+            binding.contentContainer.updatePadding(
+                left = insets.systemWindowInsetLeft + initialState.margins.left,
```
takahirom: Thank you!
I think we should use paddings for updating padding. 👀 
```
initialState.**paddings**.left
```

fornewid: 🤦‍♂It's a mistake. I changed the wrong codes. :)

android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
```diff
@@ -92,8 +94,21 @@ class MainActivity : DaggerAppCompatActivity() {
         setupNavigation()
         setupStatusBarColors()
 
-        binding.drawerLayout.doOnApplyWindowInsets { _, insets, _ ->
+        binding.drawerLayout.doOnApplyWindowInsets { _, insets, initialState ->
             binding.drawerLayout.setChildInsetsWorkAround(insets)
+            binding.contentContainer.updatePadding(
```
takahirom: Probably you should add these callbacks 🙇 🙇 
Because of initialState is that `view.doOnApplyWindowInsets` view margin or paddings

```
binding. contentContainer.doOnApplyWindowInsets { _, insets, initialState ->
```

```
binding. navView.doOnApplyWindowInsets { _, insets, initialState ->
```

android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
```diff
@@ -95,11 +97,28 @@ class MainActivity : DaggerAppCompatActivity() {
...
             }
         }
+        binding.navView.doOnApplyWindowInsets { _, insets, initialState ->
+            binding.navView.apply {
+                // On seascape mode only, nav bar is overlapped with DrawerLayout.
+                // So set left padding and reset width.
+                val leftSpace = insets.systemWindowInsetLeft + initialState.paddings.left
+                updatePadding(left = leftSpace)
+                updateLayoutParams<ViewGroup.MarginLayoutParams> {
+                    width = resources.getDimensionPixelSize(R.dimen.nav_drawer_width) + leftSpace
```
yurihondo: Hi,
Thank you for helpful cooperation.
Sorry for the numerous request but 
It is not very appropriate to fix the width of drawer when applying to various screen sizes. 
How about this?
```
                updateLayoutParams<ViewGroup.LayoutParams> {
                    if (getWidth() > 0) { // Ignore when drawer is not created
                        width = measuredWidth + leftSpace
                    }
                }
```

Could you consider this 🙏 ?

fornewid: Good! Thanks for advices.

fornewid:
:satisfied: :tada:

yurihondo:
@takahirom 
May I merge this change?

takahirom:
Thanks for working a lot! 💯

PR #266, opened 2020-01-15
Expand front layer when back button pressed

### LINKED ISSUE #126: If you back gestured or pressed when front layer is inactive, activate it
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)
- when user complete filtering session by tags, some users may do back action.
    - If user viewing session screen with no back stacks, app will close.
- `inactive front layer` is `BottomSheet expanded`
- `active front layer` is `BottomSheet collapsed`

 ## Links	

- https://material.io/components/backdrop/#behavior

gen0083:
fix issue #126

## Issue
- close #126 

## Overview (Required)
- handle back button pressed with OnBackPressedDispatcher
- handle event only if filter appeared (front layer collapsed) because if callback always enabled we cannot back to home 😢 
- it's my first time to use OnBackPressedDispatcher so I'm glad to tell me if you know how to pass back-button pressed event to default behavior

## Links
- https://developer.android.com/guide/navigation/navigation-custom-back

## Screenshot
Before | After
:--: | :--:
before | after

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
@@ -4,6 +4,8 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.activity.OnBackPressedCallback
+import androidx.activity.addCallback
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.view.children
 import androidx.databinding.DataBindingUtil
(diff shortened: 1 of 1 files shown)


takahirom:
👀

takahirom:
Thanks👍
When app collapse 2 tabs.We can not close app.👀

gen0083:
got it. i'll fix it 💪

gen0083:
update, backPressedCallback enabled only if fragment is shown (meant onResume) and front layer is collapsed (meant filter is shown).

takahirom:
It's working nicely! The code is a bit complicated, so I'm thinking about making it easy.

gen0083:
yah, i rethink too. 😄

takahirom:
Thanks! It seems better. But this has bug.  😭 
show filter → system task list -> droidkaigi -> back (not handled  )

gen0083:
😇 
it seems to be needed callback enable on onResume.
I'll fix it.

gen0083:
fixed 🙇 

by the way, how do you think about this behavior.
when drawer is opened and filter is shown, it closed filter on back button pressed.
it looks a little bit strange.
we should close drawer with back button same with this issue?

with_drawer

takahirom:
Can you open an issue? 🙇

gen0083:
sure 😄

PR #268, opened 2020-01-15
Fix wrong case of method's name

wada811:
## Issue
- N/A

## Overview (Required)
- Fix wrong case
    - LifeCycle -> Lifecycle
- Fix method's name (unify with other method's name)
    - providesLifecycleLiveData -> providesLifecycleOwnerLiveData

## Links
- N/A

## Screenshot
- N/A


takahirom:
I understand why I felt something wrong.😂

PR #278, opened 2020-01-15
Try to create universal apk

### LINKED ISSUE #217: Fix crash when opening contributor and settings
## Kind (Required)

- Other

 ## Overview (Required)

-  Probably we should upload aab instead of apk to deploy gate 🚀 

 ## Links	

-

takahirom:
## Issue
- close #217

## Overview (Required)
- Create universal apk for fixing issue
- 🙏 

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/.travis/android/script.bash
+++ b/.travis/android/script.bash
@@ -10,7 +10,16 @@ die() {
   exit 1
 }
 
-./gradlew android-base:assembleRelease -x android-base:lintVitalRelease
+./gradlew android-base:bundleRelease -x android-base:lintVitalRelease
+
+readonly aab_path="$(find android-base -name '*.aab' | head -1)"
+java -jar bundletool-all.jar build-apks \
+  --mode=universal\
+  --ks=$RELEASE_KEYSTORE_STORE_PASSWORD\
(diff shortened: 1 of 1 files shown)


ypresto:
💭 IMO 62MB is too big to be directly placed to Git 🙈

takahirom:
The jar is built by hand for using the new R8 version. 
Where do you know any good place to upload?

ypresto:
I personally prefer to use `Releases` feature of GitHub to place build artifacts. 😃

(It's hard to say, but, you can `git rm` but you can't actually remove 62MB file object from the history without force-push. 🙈)

takahirom:
Hmm. The jar is not the DroidKaigi app artifact. 🤔 
And without this, the app will crash 😭

ypresto:
Ahh I have just understood for what you are using it. It's https://github.com/google/bundletool/releases , and it's only used in Travis CI?

My recommendation for CI is to download build dependency at each build and cache it if necessary (same as maven deps).

takahirom:
The Jar is built by hand. So the jar is not uploaded.
Please see this issue 🙇 
https://github.com/DroidKaigi/conference-app-2020/issues/217#issuecomment-574545550

PR #284, opened 2020-01-15
Fix blink when favorite

### LINKED ISSUE #202: Blink time when favorite
## Kind (Required)

- Improvement

 ## Overview (Required)

- ⏲ 
like_time

 ## Links	

-

yt8492:
## Issue
- close #202 

## Overview (Required)
- use forEachIndexed
- remove `lastStartTimeText` and get last session each loop

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
@@ -5,7 +5,7 @@ import android.content.res.Resources
 import android.graphics.Canvas
 import android.graphics.Paint
 import android.view.View
-import androidx.core.view.forEach
+import androidx.core.view.forEachIndexed
 import androidx.recyclerview.widget.RecyclerView
 import com.xwray.groupie.GroupAdapter
 import io.github.droidkaigi.confsched2020.ext.getThemeColor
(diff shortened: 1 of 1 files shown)


yurihondo:
👀

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -38,18 +38,17 @@ class SessionsItemDecoration(
...
+        parent.forEachIndexed { index, view ->
             val position = parent.getChildAdapterPosition(view)
             if (position == RecyclerView.NO_POSITION) return
 
             val sessionItem = adapter.getItem(position) as SessionItem
             val startTimeText = calcTimeText(position, view)
 
-            lastStartTimeText?.let {
-                if (sessionItem.startSessionTime() == it.value) return@forEach
+            if (index > 0) {
```
yt8492: First, I tried
```
val layoutManager = parent.layoutManager as LinearLayoutManager
val topVisibleItemPosition = layoutManager.findFirstVisibleItemPosition()
if (position > topVisibleItemPosition) {
    val lastSessionItem = adapter.getItem(position - 1) as SessionItem
    if (sessionItem.startSessionTime() == lastSessionItem.startSessionTime()) return@forEach
}
```
but, first element of forEach and element at topVisibleItemPosition are not same, so flickering occured when scrolling. So I use `forEachIndexed` and check top element using index.

PR #291, opened 2020-01-16
Current session scroll

### LINKED ISSUE #150: Show current time session when loading session first in Fragment
## Kind (Required)

- Improvement

 ## Overview (Required)

- We want to show the current time session when the conference starts 📜 
- We want to scroll to the current session when first loading.
- I think we can add ViewModel property shouldScrollCurrentTime: LiveData<Boolean> = MutableLiveData(true). And when scroll to the current time, we can change that LiveData value

surajsau:
## Issue
- close #150 

## Overview (Required)
- added `val currentSessionMap = MutableLiveData<Map<SessionPage.Day, Int>>()` which is the map of the given Day and the current ongoing sessions index in the list of that day
- In the **block** of combine in `SessionsViewModel` I've added a first check of `isFirstCall = current == UiModel.EMPTY`. This is to check whether the block being executed belongs to the very first call.
- only if it is the very first call, we map the **first item in the list of sessions of a given day with `isOnGoing == true`** to the given day and set it to the LiveData
- triggering the liveData does `sessionRecycler.layoutManager.scrollToPositionWithOffset(position, 0)` to put that topic as the very first time shown in the list

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/viewmodel/SessionsViewModel.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/viewmodel/SessionsViewModel.kt
@@ -2,6 +2,7 @@ package io.github.droidkaigi.confsched2020.session.ui.viewmodel
 
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.Transformations
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.asLiveData
 import androidx.lifecycle.liveData
@@ -38,11 +39,10 @@ class SessionsViewModel @Inject constructor(
         val dayToSessionsMap: Map<SessionPage.Day, List<Session>>,
         val favoritedSessions: List<Session>,
         val filters: Filters,
-        val allFilters: Filters,
-        val showCurrentSessionButton: Boolean
+        val allFilters: Filters
     ) {
         companion object {
-            val EMPTY = UiModel(true, null, mapOf(), listOf(), Filters(), Filters(), false)
+            val EMPTY = UiModel(true, null, mapOf(), listOf(), Filters(), Filters())
         }
     }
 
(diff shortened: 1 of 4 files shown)


surajsau:
Currently working on the test cases for my changes alone

takahirom:
👀

surajsau:
@takahirom Someone has already taken up the test cases task for sessions list screen. Should I'll work on test cases once that is completed?

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/viewmodel/SessionsViewModel.kt
```diff
@@ -63,6 +64,9 @@ class SessionsViewModel @Inject constructor(
 
     private val filterLiveData: MutableLiveData<Filters> = MutableLiveData(Filters())
 
+    // map of the given Day and the current ongoing sessions index in the list of that day
+    val currentSessionMap = MutableLiveData<Map<SessionPage.Day, Int>>()
```
takahirom: Thank you for PR 📜 
Due to the complexity of the data flow, I don't want to publish LiveData for Fragment other than UiModel.

takahirom: Let me think a bit about what to do.
If you have any ideas, please try to implement them.

surajsau:
Sorry for tagging the wrong issue in the commit message. @takahirom can you check this [commit](https://github.com/DroidKaigi/conference-app-2020/pull/291/commits/62a938d29f8259fbd8447060707a6e79d2a0107d) I've moved the map into UiModel itself. This way no other livedata are published

takahirom:
How about like this?
https://github.com/DroidKaigi/conference-app-2020/compare/current-time-scroll?expand=1

surajsau:
Got the intention. The only issue I see is the function call on scroll completion will update shoudlScroll LiveData which will again trigger combine() which may, in a corner case, again trigger the scroll completion function.

That was my only worry with this implementation. I’ll work on this perspective.

takahirom:
Please let me know if you have any problems. 🙇

surajsau:
I tried doing a different approach inspired by your suggestion from [this](https://github.com/DroidKaigi/conference-app-2020/compare/current-time-scroll?expand=1). But I ended up almost doing it like yours. I think we can go ahead with your implementation itself. ☺️

Currently unable to think of a better way of controlling the `combine` based on a `LiveData`

takahirom:
Thanks!

PR #292, opened 2020-01-16
Add year and organization name

Luecy1:
## Issue
- None

## Overview (Required)
We should fill out year and name in  License file.

## Links
I referenced ↓
- https://github.com/JetBrains/kotlin/blob/master/license/LICENSE.txt#L190

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2020 DroidKaigi and contributors.
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2020 DroidKaigi and contributors.
```
takahirom: Thanks for the suggestions!
We discussed a little.  The "DroidKaigi committee" seems to be better to protect contributors from risk.

Luecy1: Thank you for comment.
I'll fix it.

takahirom:
LGTM! 📜

PR #294, opened 2020-01-16
Apply AutoCleardValue to Fragment member variables

### LINKED ISSUE #153: Fix fragment view leaks

TomoyaShibata:
## Issue
- close #153

## Overview (Required)

- References to view, adapter, and more remain unintentionally when add a fragment to a backstack.
- Applied [AutoClearedValue](https://github.com/android/architecture-components-samples/blob/9826b59956eb93c9e627bdf16a19a1c8bc28ce14/GithubBrowserSample/app/src/main/java/com/android/example/github/util/AutoClearedValue.kt) solved this issue.
  - AutoClearedValue is a delegated property that implements ReadWriteProperty. Set null in the onDestroy life cycle of viewLifecycleOwner.lifecycle.

## Links

- https://github.com/DroidKaigi/conference-app-2020/pull/230

### DIFF
--- a/corecomponent/androidcomponent/src/main/java/io/github/droidkaigi/confsched2020/util/AutoClearedValue.kt
+++ b/corecomponent/androidcomponent/src/main/java/io/github/droidkaigi/confsched2020/util/AutoClearedValue.kt
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2018 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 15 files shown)


takahirom:
👀

takahirom:
Thanks! Perhaps it is not leaking in ProgressTimeLatch, but it seems better to null everything put in the field in onCreateView in this way. 👍 
You may want to have an Adapter etc. in the field as usual.
It seems good to adapt if there is a library that can solve this problem.
Thanks for trying.

For now, let's merge this PR

PR #307, opened 2020-01-16
Make sessions available for calendaring

### LINKED ISSUE #163: Make sessions available for calendaring
## Kind (Required)

- Improvement

 ## Overview (Required)

- It is more convenient to be able to register to Google Calendar.
image

 ## Links	

-

furusin:
## Issue
- close #163 

## Overview (Required)
- adding a function for create a Calendar item.

## Screenshot
|before|after|
|:--:|:--:|
|<image src="https://user-images.githubusercontent.com/2215210/72532589-0f97d200-38b7-11ea-998d-fe82534c60f2.png" width="300px" />|<image src="https://user-images.githubusercontent.com/2215210/72532663-2b02dd00-38b7-11ea-8c34-6574e096d4de.png" width="300px" />|


jmatsu:
👀

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
```diff
@@ -92,7 +92,27 @@ class SessionDetailFragment : DaggerFragment() {
             }
 
         binding.bottomAppBar.setOnMenuItemClickListener {
-            handleNavigation(it.itemId)
+            when (it.itemId) {
```
jmatsu: Local variable declaration in this `when` subject would be better.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
```diff
@@ -92,7 +92,27 @@ class SessionDetailFragment : DaggerFragment() {
...
+                R.id.session_share -> {
+                    // do something
+                }
+                R.id.floormap -> {
+                    handleNavigation(it.itemId)
+                }
+                R.id.session_calendar -> {
+                    val session = binding.session ?: return@setOnMenuItemClickListener false
```
jmatsu: `true` sounds better and consume a click event if returning a result somehow.

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
```diff
@@ -35,4 +36,20 @@ class SystemViewModel @Inject constructor(
         // launch a Custom Tabs Activity
         customTabsIntent.launchUrl(activity, Uri.parse(url))
     }
+
+
+    fun openCalendar(
```
jmatsu: The aim is not to open a calendar but saving a plan to a calendar, right? then `saveEventToCalendar` would be nicer.

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
```diff
@@ -35,4 +36,20 @@ class SystemViewModel @Inject constructor(
...
+        title: String,
+        location: String,
+        startUnixMillis: Long,
```
jmatsu: `DateTime` is preferable to avoid a time-unit mismatch if the class is in the classpath.

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
```diff
@@ -35,4 +36,20 @@ class SystemViewModel @Inject constructor(
...
+        endUnixMillis: Long
+    ) {
+        val intent = Intent(Intent.ACTION_INSERT)
+            .setData(CalendarContract.Events.CONTENT_URI)
+            .putExtra(CalendarContract.EXTRA_EVENT_BEGIN_TIME, startUnixMillis)
+            .putExtra(CalendarContract.EXTRA_EVENT_END_TIME, endUnixMillis)
+            .putExtra(CalendarContract.Events.TITLE, "DroidKaigi2020: $title")
+            .putExtra(CalendarContract.Events.EVENT_LOCATION, location)
+        activity.startActivity(intent)
```
jmatsu: @takahirom 
I think we should handle it somehow if a device does not have any Calendar application. However, we may need to consider the design of handling such an exception. Can I make it a future work and accept this change as it is now?

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
```diff
@@ -35,4 +36,20 @@ class SystemViewModel @Inject constructor(
...
     }
+
+
+    fun openCalendar(
+        title: String,
+        location: String,
+        startUnixMillis: Long,
+        endUnixMillis: Long
+    ) {
+        val intent = Intent(Intent.ACTION_INSERT)
```
jmatsu: 👍

feature/session/src/main/res/values/strings.xml
```diff
@@ -4,6 +4,7 @@
     <string name="start_filter">Filter</string>
     <string name="intended_audience_title">Intended audience</string>
     <string name="session_share" translatable="false">Share</string>
+    <string name="session_register" translatable="false">Register</string>
```
jmatsu: `save_event_to_calendar` and `Save to Calendar` sound more natural.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
```diff
@@ -92,7 +92,27 @@ class SessionDetailFragment : DaggerFragment() {
...
+                    systemViewModel.openCalendar(
```
jmatsu: Could you please use named arguments for here?

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
```diff
@@ -92,7 +92,27 @@ class SessionDetailFragment : DaggerFragment() {
...
+                        session.title.getByLang(defaultLang()),
+                        session.room.name.getByLang(defaultLang()),
```
jmatsu: A room name as a part of event information makes sense. So cool.
Following your idea, I'm thinking attendees would be happy if they could know which room they should enter at a glance.
For example, it could be achieved if a calendar event title is `[<room name>] <session title>`. (in this case, the location column is not required). What do you feel this feedback?

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
```diff
@@ -92,7 +92,27 @@ class SessionDetailFragment : DaggerFragment() {
             }
 
         binding.bottomAppBar.setOnMenuItemClickListener {
-            handleNavigation(it.itemId)
+            when (it.itemId) {
+                R.id.session_share -> {
+                    // do something
+                }
+                R.id.floormap -> {
```
jmatsu: seems to be unnecessary branch?

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
takahirom: We are removing activity from SystemViewModel property. Please add activity parameter to openCalendar.
https://github.com/DroidKaigi/conference-app-2020/issues/323

takahirom: It is better to try-catch when starting implicit Intent. 👍

furusin: Sure.
I'll marge this PR after that. (Must be conflicted)

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
furusin: Thank you for your feedback.
Your idea sounds fine, but I am wondering that the title of calendar could be too long like below:
> [Hall A] Jetpack時代…

Why don't we just try add the room name at this time, and then touch and try after room name attached?
Fortunately, We can change for anyway.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
furusin: updated with this issue 3d7dfc9

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
furusin: updated with this issue e4e55ca

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
furusin: updated with this issue d4786eb
I thought `saveEventToGoogleCalendar` is better.

feature/session/src/main/res/values/strings.xml
furusin: updated with this issue f4a19f5

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
furusin: updated with this issue f96e2ba

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
jmatsu: > Why don't we just try add the room name at this time, and then touch and try after room name attached?
> Fortunately, We can change for anyway.

What will we try? and what for change? Could you please rephrase it, sorry?

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
jmatsu: I'm just curious so let me ask a question on the new name. Does it mean this code work with only Google calendar app?

takahirom:
@jmatsu Can you check this? 🙇

jmatsu:
I already reviewed and asked several questions about the latest changes. And also, it looks some pending reviews are available.

https://github.com/DroidKaigi/conference-app-2020/pull/307#discussion_r367452932
https://github.com/DroidKaigi/conference-app-2020/pull/307#discussion_r367444207
https://github.com/DroidKaigi/conference-app-2020/pull/307#discussion_r367740651
https://github.com/DroidKaigi/conference-app-2020/pull/307/files#diff-2b4a7fcb147ae35e0c939da24e583057R41

furusin:
I thought that was pending for review 🙇 
I'll update.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
furusin: 「まずはルーム名を追加して試してみて、ダメだったら変更するようにしませんか？」と言いたかったです

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
furusin: Sure, I'll remove.

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
furusin: We should use `DateFormat` , you mean?
I couldn't figure out what I need to change for, sorry.

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
furusin: Oops, I thought it was explicit intent but wasn't. I'll remove `Google` .

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
jmatsu: Gotcha. I'm fine.

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
jmatsu: JFYI, the FQDN of `DateTime` is `com.soywiz.klock.DateTime`

furusin:
@jmatsu 
i've updated. Please check them out.

feature/session/src/main/res/menu/menu_session_detail_bottomappbar.xml
```diff
@@ -1,21 +1,24 @@
 <?xml version="1.0" encoding="utf-8"?>
-<menu
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    >
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
```
jmatsu: Sorry, I overlooked this point. Could you please format this file?

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
```diff
@@ -13,4 +17,20 @@ class SystemViewModel @Inject constructor() : ViewModel() {
     fun onError(error: AppError) {
         mutableErrorLiveData.value = error
     }
+
+    fun sendEventToCalendar(
+        context: Context,
```
jmatsu: This should be *activity* as @takahirom has requested. This can crash applications if any non-activity context has been passed.  And also, in this case, there is no reason to create a new task by adding `FLAG_ACTIVITY_NEW_TASK` to an intent that will be created so only changing *Context* to *Activity* is acceptable.

feature/system/src/main/java/io/github/droidkaigi/confsched2020/system/ui/viewmodel/SystemViewModel.kt
```diff
@@ -13,4 +19,24 @@ class SystemViewModel @Inject constructor() : ViewModel() {
...
+        val intent = Intent(Intent.ACTION_INSERT)
+            .setData(CalendarContract.Events.CONTENT_URI)
+            .putExtra(CalendarContract.EXTRA_EVENT_BEGIN_TIME, startDateTime.unixMillisLong)
+            .putExtra(CalendarContract.EXTRA_EVENT_END_TIME, endDateTime.unixMillisLong)
+            .putExtra(CalendarContract.Events.TITLE, "[$location] $title")
+            .putExtra(CalendarContract.Events.EVENT_LOCATION, location)
+        try {
+            activity.startActivity(intent)
+        } catch (e: Exception) {
+            Timber.debug(e) { "Fail startActivity" }
```
jmatsu: Got it. That's fine with me.

PR #331, opened 2020-01-17
Fix image blink when we favorite session

### LINKED ISSUE #288: Blink time and image when favorite status change
## Kind (Required)
- Improvement

 ## Overview (Required)
When session's favorite status is changed, `SessionItem` update and call bind method. (equals returns false)

 ## Links	

-

rmakiyama:
## Issue
- close #288 

## Overview (Required)
- fix issue as using payload

## Links
- [Groupie sample](https://github.com/lisawray/groupie/blob/8b79fa16093a9cb8e7efeefc68c47713157e1422/example-databinding/src/main/java/com/xwray/groupie/example/databinding/item/HeartCardItem.java#L73)

## Screenshot
Before | After
:--: | :--:
 |


jmatsu:
👀

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
```diff
@@ -73,6 +69,37 @@ class SessionItem @AssistedInject constructor(
...
 
+    override fun bind(
+        viewBinding: ItemSessionBinding,
+        position: Int,
+        payloads: MutableList<Any>
+    ) {
+        if (payloads.isEmpty()) {
+            bind(viewBinding, position)
+        } else {
+            payloads.forEach { payload ->
```
jmatsu: *payloads* would consist of `notifyItemChanged`'s payload and `notifyItemRangeChanged` payload. That's why the type of payloads is List. 

So `payloads.distinct` should be called in this design. Because binding would be called multiple times if multiple `ItemPayload.FavoritePayload` are included.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
```diff
@@ -73,6 +69,37 @@ class SessionItem @AssistedInject constructor(
         viewBinding.speakers.bindSpeaker()
     }
 
+    override fun bind(
+        viewBinding: ItemSessionBinding,
+        position: Int,
+        payloads: MutableList<Any>
+    ) {
+        if (payloads.isEmpty()) {
+            bind(viewBinding, position)
```
jmatsu: 👍

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
```diff
@@ -162,18 +189,35 @@ class SessionItem @AssistedInject constructor(
     }
 
     override fun equals(other: Any?): Boolean {
-        return isSameContents(other)
+        if (other !is SessionItem) return false
+        return when {
+            isSameContents(other) -> true
+            isChangeFavorited(other) -> {
+                other.notifyChanged(ItemPayload.FavoritePayload(session.isFavorited))
```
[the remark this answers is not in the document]
rmakiyama: Thank you!! I always wanted a better way and your advice is exactly way!! 👏

rmakiyama:
@jmatsu 
I fixed. please review again 🙏

rmakiyama:
I learned so much! Thanks!!!!!

Issue #133, opened 2020-01-11
Key board is not hidden when back from search by button


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- If you move screen while the keyboard is displayed, the keyboard will not be hidden.
- Hide keyboard when moved.

![keyboard_not_hidden_when_back](https://user-images.githubusercontent.com/1386930/72201433-3dd07880-3497-11ea-85e7-3ae806e4a7e0.gif)

 ## Links	

-

kumamotone:
🙋‍♂

roana0229:
@kumamotone Thanks, assigned!

PR #339, opened 2020-01-17
Hide keyboard when moved to other Fragment

kumamotone:
## issue

close https://github.com/DroidKaigi/conference-app-2020/issues/133

## Overview (Required)

Hide keyboard when moved to other Fragment.

I used `onDestroyView()` of Fragment because implementation is short and simple.
It seems to work but I'm not sure (because it's first time to use this way honestly). Please check whether appropriate or not 🙏


roana0229:
👀

roana0229:
I'm first time to use this way too.
I confirmed that "onDestroy" and "onDetach" was called after "onDestroyView" ✅ 

It looks like good. 👍

PR #351, opened 2020-01-17
#349 ViewModel values ​​except UiModel should be inaccessible from outside

### LINKED ISSUE #349: ViewModel values ​​except UiModel should be inaccessible from outside
## Kind (Required)

- Improvement

 ## Overview (Required)

- Sometimes, we can access public instances of viewModel. But many of them should be private.
- Please try look into various viewModels for refactoring!

Pluu:
## Issue
- close #349

## Overview (Required)
- Modify accessor of public instance to private


takahirom:
LGTM 🚮

PR #354, opened 2020-01-17
Fix bottom app bar padding is not exist

### LINKED ISSUE #62: Bottom app  bar padding is not exist on Android 5.0
image

cheddar-dev:
## Issue
- close #62 

## Overview (Required)
- This bug occurs when the navigation bar is below the app bar.
- If navigation bar is bottom, add navigation bar height to app bar padding.

## Links
- N/A

## Screenshot
| | Before | After
| :--: | :--: | :--:
| Nexus5X Android5.0 |  | 
| Nexus9 Android7.1.1 |  |


feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
```diff
@@ -109,6 +112,16 @@ class SessionDetailFragment : DaggerFragment() {
         binding.bottomAppBar.setOnMenuItemClickListener {
             handleNavigation(it.itemId)
         }
+
+        val windowManager = context?.getSystemService(Context.WINDOW_SERVICE) as? WindowManager
```
takahirom: Thanks! I think readable code by using variables! 👍 
By the way, Can we use inserter?

cheddar-dev: sorry, I'm not good at English. 😭 
Do you mean code like this?
`val display = (context?.getSystemService(Context.WINDOW_SERVICE) as? WindowManager)?.defaultDisplay`

takahirom: Sorry  🙇 
Other screens use `insetter.doOnApplyWindowInsets`
https://github.com/DroidKaigi/conference-app-2020/blob/d64a1ce19837c413e5cdc0ad14d5fe202a4c0213/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt#L99

cheddar-dev: Thank you!! I'll fix 👀

takahirom: If this doesn't work, you may adopt your plan.

cheddar-dev: I tried using `OnApplyWindowInsets` on `onViewCreated` and `onCreateView`, but it wasn't working. 
I'm sorry, but the problem difficult for me.
Could you decide whether to close or merge this PR? sorry... :pray:

cheddar-dev:
@takahirom I fixed conflict and fix jmatsu-bot's point out. Please check and merge 🙏

takahirom:
Maybe this idea will be applied to the whole app, so let me think a bit.  🙇
I want to know why insetter.doOnApplyWindowInsets was no good.  👀
1. The callback was not called.  
2. Padding was called with 0.

cheddar-dev:
`doOnApplyWindowInsets` is called, and updatePadding is called. 
But, It looks like padding is not set.

I'm not sure, but padding can set only before onViewCreated...?
But, Other screen is good. Umm... 🤔 

```
binding.bottomAppBar.doOnApplyWindowInsets { _, insets, initialState ->
           // This callback is called.
           // And, (insets.systemWindowInsetBottom + initialState.paddings.bottom) == (realScreenSize.y - appUsableSize.y).
           // But, It looks like padding is not set.
            binding.bottomAppBar.updatePadding(
                bottom = insets.systemWindowInsetBottom + initialState.paddings.bottom
            )
        }
```

cheddar-dev:
@takahirom I understand you are very busy, but how should I do this PR?
Could you please confirm https://github.com/DroidKaigi/conference-app-2020/pull/354#issuecomment-576279380 🙏

takahirom:
Sorry for late 👀

takahirom:
The code here is likely related to the Behavior of the CoordinatorLayout 👀

takahirom:
The position of the bottom sheet seems to change depending on the position of the Fab (BottomAppBar.addFabAnimationListeners) 👀

yurihondo:
👀

yurihondo:
@cheddar-dev 
Hi, 
Sorry for late reply to your comment.
I've investigate this issue and found one of a solution.
So, I take over the resolving of this issue :D
Sorry for the too late support, however thanks to you we have noticed the difficulty of this issue.
Thank you for your cooperation.

yurihondo:
The following PR has been taken over.
#741

PR #367, opened 2020-01-18
Add method isEdgeToEdgeEnabled

rwadada:
## Issue
- none

## Overview (Required)
- Added 48dp space to sessionSheetBehavior.peekHeight only when gesture navigation is enabled.
(note: May not work properly with emulator)

## Links
- none

## Screenshot
　 | Before | After
:--: | :--: | :--:
Gesture navigation |  | 
2button navigation |  | 
3button navigation |  |


takahirom:
Thank you!It has become much easier to use!! 
But, There are some design concerns. 
Could you create another issue like the one below?(You can use this image.)
It seems good to use MotionLayout to scroll the line down as you scroll, what do you think?
image

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
```diff
@@ -194,6 +201,23 @@ class SessionsFragment : DaggerFragment() {
...
+    /**
+     * judge gesture navigation is enabled
+     * https://android.googlesource.com/platform/packages/apps/Settings.git/+/refs/heads/master/src/com/android/settings/gestures/SystemNavigationPreferenceController.java#97
+     *
+     * If configNavBarInteractionMode is equal to "2", it means gesture navigation
+     * https://android.googlesource.com/platform/frameworks/base/+/refs/heads/android10-mainline-release/core/java/android/view/WindowManagerPolicyConstants.java#60
+     * */
+    private fun isEdgeToEdgeEnabled(): Boolean {
+        val configNavBarInteractionMode = Resources.getSystem().getIdentifier(
+            "config_navBarInteractionMode",
```
takahirom: Are you a hacker? 😄

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
```diff
@@ -194,6 +201,23 @@ class SessionsFragment : DaggerFragment() {
...
+            "integer",
+            "android"
+        )
+        return (Build.VERSION.SDK_INT > Build.VERSION_CODES.P &&
```
takahirom: I prefer the early return a bit. 👀 

```
private fun isEdgeToEdgeEnabled(): Boolean {
  if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) return false
```

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
```diff
@@ -98,8 +101,12 @@ class SessionsFragment : DaggerFragment() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         val initialPeekHeight = sessionSheetBehavior.peekHeight
+        val gestureNavigationBottomSpace =
+            if (isEdgeToEdgeEnabled()) (48 * resources.displayMetrics.density).toInt() else 0
```
takahirom: Can you use a dimens resource?

rwadada: Yes, that is better. Modify to get from dimen resource.

rwadada:
> It seems good to use MotionLayout to scroll the line down as you scroll, what do you think?

I think it's very nice! I will create an issue about this 👍 
Please let me know if I have any misunderstandings about my created issue.

takahirom:
Thank you for creating the issue and hack 😄 !

PR #371, opened 2020-01-18
Updated to use navigation component by Kotlin

FujiKinaga:
## Issue
- N/A

## Overview (Required)
- Updated versions about navigation component
- Deleted unused libraries
    - If full kotlin project, we can use only ktx libraries
- To use `safe args kotlin`, generated codes are changed to kotlin codes.
- Try to use navigation method on click event for navigating to staffs
    - What I was doing was the same as before...

## Links
- https://developer.android.com/guide/navigation/navigation-getting-started
- https://developer.android.com/guide/navigation/navigation-navigate

## Screenshot
Generated codes

by Java | by Kotlin
:--: | :--:
 |

### DIFF
--- a/corecomponent/androidcomponent/build.gradle
+++ b/corecomponent/androidcomponent/build.gradle
@@ -2,7 +2,7 @@ import dependencies.Dep
 
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
-apply plugin: 'androidx.navigation.safeargs'
+apply plugin: 'androidx.navigation.safeargs.kotlin'
 apply plugin: 'kotlin-kapt'
 
 apply from: rootProject.file('gradle/android.gradle')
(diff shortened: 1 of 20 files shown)


corecomponent/androidcomponent/build.gradle
```diff
@@ -37,7 +36,6 @@ dependencies {
     implementation Dep.Firebase.auth
 
     api Dep.AndroidX.lifecycleExtensions
-    api Dep.AndroidX.Navigation.runtimeKtx
```
FujiKinaga: It seems to duplicate🙏

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
```diff
@@ -242,9 +242,7 @@ class SessionsFragment : DaggerFragment() {
         val fragment: Fragment = when (tab) {
             is SessionPage.Day -> {
                 BottomSheetDaySessionsFragment.newInstance(
-                    BottomSheetDaySessionsFragmentArgs
-                        .Builder(tab.day)
```
takahirom: Great 👍

Issue #155, opened 2020-01-12
Add ViewModel test for session search screen


takahirom:
## Overview (Required)
The title says everything. 💯 
You can implement it in a similar way to SessionDetailViewModelTest.
https://github.com/DroidKaigi/conference-app-2020/blob/8691c73ccfaa7ba626752433d7b97d744a2e679f/feature/session/src/test/java/io/github/droidkaigi/confsched2020/session/ui/viewmodel/SessionDetailViewModelTest.kt

## Links
https://github.com/DroidKaigi/conference-app-2020/blob/master/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/viewmodel/SearchSessionsViewModel.kt

sdsd08013:
🙋‍♂

mhidaka:
Assigned 👍

FujiKinaga:
Fixed by #373 💯

PR #373, opened 2020-01-18
Add ViewModel test for session search screen 

sdsd08013:
## Issue
https://github.com/DroidKaigi/conference-app-2020/issues/155

## Overview (Required)
- Add SearchSessionsViewModel Unit Test.

## Links
- N/A

## Screenshot
- N/A


sdsd08013:
Question

Is there any reason ServerSession description is excluded from find conditions?
https://github.com/DroidKaigi/conference-app-2020/blob/master/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/SessionContents.kt#L40

I think it's better to include this for UX.

takahirom:
Thanks for suggestions!
There are not many service sessions, many of which are related to the whole, such as lunch and keynotes.
Please let us know if you have any opinions.

sdsd08013:
I understand.
It makes sense!

PR #391, opened 2020-01-18
add bottom margin to timeline filter.

### LINKED ISSUE #384: There is no bottom padding at Timeline Filter
## Kind (Required)

- Improvement

 ## Overview (Required)

- It is necessary 40dp padding between chip and bottom sheet in Timeline Filter.

## Links

- https://www.figma.com/file/4r9becvhDy3GfXaXex8E8d/App?node-id=11%3A625

## Image
| Figma | App |
|---|---|
| image | screenshot |

tsumuchan:
## Issue
- close #384

## Overview (Required)
- add bottom margin to timeline filter.

## Links
- Figma: https://www.figma.com/file/4r9becvhDy3GfXaXex8E8d/App?node-id=11%3A625

## Screenshot
Before | After
:--: | :--:
|


takahirom:
Thanks!
Can we use ConstraintLayout paddingBottom and clipToPadding =false?
Because If we add an item, we should add the same constraint 👀

tsumuchan:
Thank you review my PR :pray:

> Because If we add an item, we should add the same constraint 👀

Yes, You’re right!! I fixed it.

takahirom:
Thanks! LGTM 🙏

Issue #118, opened 2020-01-11
Filters are not exists on Android 5.0


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- 

![image](https://user-images.githubusercontent.com/1386930/72200582-a109dd00-348e-11ea-8d36-e31ad91aff26.png)

gen0083:
filters are shown but need some long times or need to move day1-day2 etc...
attached screenshot on Android5.0 with emulator.
(probably same on Android 5.1.1)

![Screenshot_1579164537](https://user-images.githubusercontent.com/7608725/72508265-bf098000-3888-11ea-8652-273b688dc637.jpg)

crimsonwoods:
@takahirom Maybe, I found the workaround to fix this issue. Would you assign me?

takahirom:
Thanks Assigned 👍

PR #395, opened 2020-01-18
Make favorite icon state update right away when clicked

### LINKED ISSUE #122: Favorite is not applied on Android 5.0
## Kind (Required)

- Improvement

 ## Overview (Required)

- Probably the favorite state applied to session, but not shown.
favorite_not_applied

 ## Links	

-

eyedol:
## Issue
- close #122 

## Overview (Required)
- Make favorite icon sync updated instead of async after update call.
- Utilized icon selector to update the favorite icon state.
- Minor cosmetic clean up.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
@@ -3,6 +3,7 @@ package io.github.droidkaigi.confsched2020.session.ui.item
 import android.content.Context
 import android.view.LayoutInflater
 import android.view.ViewGroup
+import android.widget.ImageButton
 import android.widget.ImageView
 import android.widget.TextView
 import androidx.core.content.ContextCompat
@@ -61,7 +62,7 @@ class SessionItem @AssistedInject constructor(
                 favorite.isSelected = !favorite.isSelected
                 sessionsViewModel.favorite(session)
             }
-            favorite.isSelected = session.isFavorited
+            bindFavorite(session.isFavorited, favorite)
             root.setOnClickListener {
                 root.findNavController().navigate(actionSessionToSessionDetail(session.id))
             }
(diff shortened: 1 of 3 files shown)


takahirom:
👀

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
```diff
@@ -52,52 +51,22 @@ class SessionItem @AssistedInject constructor(
...
-
-    override fun bind(
-        viewBinding: ItemSessionBinding,
-        position: Int,
-        payloads: MutableList<Any>
-    ) {
-        if (payloads.isEmpty()) {
-            bind(viewBinding, position)
-        } else {
-            payloads.distinct().forEach { payload ->
```
takahirom: Thank you for investigate!!! Is this code not needed? 👀 
Sure, this code is to prevent the session from blinking.

takahirom:
This problem is quite difficult. 😭 
Maybe I think this may fix this, but what do you think?
https://github.com/DroidKaigi/conference-app-2020/pull/398

eyedol:
Lemme try it

eyedol:
@takahirom that worked and fixed issue #118

eyedol:
Check my [comment](https://github.com/DroidKaigi/conference-app-2020/pull/398/files#r368240905) Let me know what to do. There's a fundamental flaw somewhere 😠

eyedol:
Also, my current workaround shouldn't affect anything. wdyt?

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
takahirom: @eyedol 
Can you check this?? 🙏 
We are overriding `override fun getChangePayload` for SessionItem. But now we are not handling that. 👀

takahirom:
Thanks!
Since it seems that it is not related to #398, favorites will be fixed in #395 separately.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionItem.kt
eyedol: > Thank you for investigating!!! Is this code not needed? 👀
> This code is to prevent the session from blinking.

Looks like it was removed from a merge conflict. Checking...

eyedol: @takahirom  fixed 🙇

feature/session/src/main/res/drawable/favorite_button_selector.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
```
takahirom: 🆒

eyedol:
Can I get a merge so I don't maintain merge conflicts? 🙏

takahirom:
Thanks!

PR #404, opened 2020-01-18
Fix options menu not shown when rotate

### LINKED ISSUE #388: Content area must be scrollable and viewable in preparation for screen rotation
## Kind (Required)

- Improvement

 ## Overview (Required)

- Sometimes, we can't view screens because of non-scrollable and non-margin.
- Please try look into various screen layout for refactoring!

 ## Screenshot	
For example, just like this.

device-2020-01-18-180302

5hyn3:
## Issue
- close #388

## Overview (Required)
- Fix bug that options menu not shown when the screen is rotated.
  - Investigate by the layout inspector, OptionsMenu was drawn. But, it's position was wrong.Call `invalidateOptionsMenu()` when `doOnApplyWindowInsets` then it was resolve.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
+++ b/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
@@ -106,6 +106,7 @@ class MainActivity : DaggerAppCompatActivity() {
             view.updateLayoutParams<ConstraintLayout.LayoutParams> {
                 topMargin = insets.systemWindowInsetTop + initialState.margins.top
             }
+            // Invalidate because option menu cannot be displayed after screen rotation
             invalidateOptionsMenu()
         }
         binding.navView.doOnApplyWindowInsets { view, insets, initialState ->
--- a/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
+++ b/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
@@ -1,9 +1,7 @@
 package io.github.droidkaigi.confsched2020
 
 import android.annotation.SuppressLint
-import android.content.res.Configuration
 import android.os.Bundle
-import android.view.Menu
 import android.view.View
 import androidx.annotation.IdRes
 import androidx.appcompat.content.res.AppCompatResources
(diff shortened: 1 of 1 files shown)


takahirom:
I googled. Is this related?
https://github.com/takahirom/conference-app-2020/blob/master/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt#L169
https://stackoverflow.com/a/29541621/4339442

5hyn3:
> Thanks!
> Since I do not know the cause, I'm afraid that I have to put this code everywhere. 👀
> I googled. Is this related?
> https://github.com/takahirom/conference-app-2020/blob/master/android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt#L169
> https://stackoverflow.com/a/29541621/4339442

Thanks for providing helpful link!

In conclusion, commenting out `setTitle (...)` did not solve the problem.
The problem seems to have occurred because did not consider that navigationBar was hidden.
Checking the coordinates of OptionMenus, it seems to exist at the position of NavigationBar.

So this problem can also be solved by commenting out the following: (But the UI collapses.)

https://github.com/takahirom/conference-app-2020/blob/master/android-base/src/main/java/io/github/droidkaigi/confsched2020/ui/widget/SystemUiManager.kt#L62

Of course, I didn't want to take any measures to break the UI, so I decided to call `invalidateOptionsMenu ()`.

takahirom:
I have a question 🙋 
There are several other View events, but why do you do this at the time of doOnApplyWindowInsets? 👀

5hyn3:
From the above information, I speculated that this problem was caused by the behavior of ApplyWindowInsets due to a change in systemUiVisibility.
So call invalidApplyMenus () in doOnApplyWindowInsets.
But I'm not very familiar with the behavior around WindowInsets.
While studying, I survey for better timing. 🙇

takahirom:
Thanks! Can you add a comment like this?

```
// Invalidate because option menu cannot be displayed after screen rotation
```

5hyn3:
OK! I add it ASAP!

5hyn3:
Comment added!

takahirom:
LGTM 👍 Thank you for investigating!!

Issue #325, opened 2020-01-16
Keep expanding state across screen rotate


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- Probably we can do it with ViewModel

- Session detail
- Announcement

ntsk:
🙋‍♂

FujiKinaga:
@ntsk 
Thanks! Assigned👍

PR #420, opened 2020-01-18
Keep expanding state across screen rotate

ntsk:
## Issue
- close https://github.com/DroidKaigi/conference-app-2020/issues/325

## Overview (Required)
- Changed to manage expanding state by ViewModel in SessionDetail and Announcement.

## Links
-


feature/announcement/src/main/java/io/github/droidkaigi/confsched2020/announcement/ui/viewmodel/AnnouncementViewModel.kt
```diff
@@ -52,7 +53,7 @@ class AnnouncementViewModel @AssistedInject constructor(
         UiModel(
             isLoading = loadState.isLoading,
             error = loadState.getErrorIfExists().toAppError(),
-            announcements = announcements,
+            announcementItems = announcements.map { announcementItemFactory.create(it) },
```
takahirom: Thanks! 👍 
Groupie Items know Android Views.
If we replace with Jetpack Compose. We can't use it.
Is there any reason you want to do this? 👀

ntsk: Thanks for the review! :smiley: 

The reason is that I thought of other ways to keep the position of the Item as follows, but I thought it was simpler than this.

```kotlin
data class UiModel(
    val isLoading: Boolean,
    val error: AppError?,
    val announcement: List<Announcement>,
    val isEmpty: Boolean,
    val showEllipsisPositon: List<Int>
)
```

However, given the migration to Jetpack Compose in the future, it is better not to have a BindableItem in a ViewModel, right?

Then I'd like to change to the method described above, any other good ideas?

takahirom: Sorry for late. I think it is good! 👍

takahirom:
👀

takahirom:
LGTM 💯

PR #422, opened 2020-01-19
Fix favorite sessions visibility

### LINKED ISSUE #418: Fix sessions are displayed after changing filters
## Kind (Required)

- Improvement

 ## Overview (Required)

- Sessions is hide when collapsed.
- But sessions are displayed after changing filters.

 ## Links	

-

umechanhika:
## Issue
- close #418 

## Overview (Required)
- Two attributes relation to visibility were set to RecyclerView.
- `isVisibleWithAnimation` is unnecessary because the animation is done with `TransitionManager`.
- Also `isEmptySessions` has no relationship to RecyclerView's visibility.

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
I think `isVisibleWithAnimation` is needed for fading empty_session image to the session list.
Can you check that?

umechanhika:
I remove `isVisibleWithAnimation` from only RecyclerView and I left it in empty_session image.
So empty_view fading is assumed that does not change.

takahirom:
It has not been confirmed, The fading is used when switching empty_session -> recyclerview. 👀

umechanhika:
I understood you said.
I checked the app operation and I notice that,,,
Before code: recyclerview is shown and empty_session image is hidden at the same time.
My fixed code: recyclerview is shown before empty_session image is hidden.

I will consider the fix again.

umechanhika:
I fixed again.
I will write explanation so please wait.
It may takes some time because it is difficult to explain in English.

takahirom:
I think the code becomes good 😄

umechanhika:
Thank you for understanding 😄

PR #431, opened 2020-01-19
Fix gradle obsolete API warnings

craf0:
## Issue
- none

## Overview (Required)
- Fix the following warning on build
```
build.gradle: API 'variant.getGenerateBuildConfig()' is obsolete and has been replaced with 'variant.getGenerateBuildConfigProvider()'.
It will be removed at the end of 2019.
For more information, see https://d.android.com/r/tools/task-configuration-avoidance.
```
## Links
- https://d.android.com/r/tools/task-configuration-avoidance

### DIFF
--- a/data/api/build.gradle
+++ b/data/api/build.gradle
@@ -10,7 +10,9 @@ apply plugin: 'kotlinx-serialization'
 apply from: rootProject.file('gradle/android.gradle')
 android.libraryVariants.all {
     // TODO replace with https://issuetracker.google.com/issues/72050365 once released.
-    it.generateBuildConfig.enabled = true
+    it.generateBuildConfigProvider.configure {
+        enabled = true
+    }
 }
 
 kotlin {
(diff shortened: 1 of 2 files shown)


data/api/build.gradle
```diff
@@ -10,7 +10,9 @@ apply plugin: 'kotlinx-serialization'
 apply from: rootProject.file('gradle/android.gradle')
 android.libraryVariants.all {
     // TODO replace with https://issuetracker.google.com/issues/72050365 once released.
-    it.generateBuildConfig.enabled = true
```
takahirom: Thanks!
Probably this gradle file load `gradle/android.gradle`. So can we delete this?

craf0: gradle/android.gradle set generateBuildConfig = false.
But data/api subprojects use BuildConfig.java (to set API_ENDPOINT).
So this line seems necessary.

takahirom: Thanks. You are correct!! 👍

PR #403, opened 2020-01-18
Use SharedElementTransition when transitioning to the Speaker screen

n-seki:
## Issue
- close #168

## Overview (Required)
- Apply SharedElementTransition to speaker image
- Also I tried to apply "return shared element" for SpeakerFragment to SessionDetailFragment transition.
  - It was difficult to apply return shared element to other transitions.

## Links
-

## Screenshot
||Before | After
|---|:--: | :--:
|Main to Speaker| | 
|Search to Speaker| | 
|Session Detail to Speaker| |

Issue #168, opened 2020-01-12
Use SharedElementTransition when transitioning to the Speaker screen.


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- The title says everything 👝 

 ## Links	

-

n-seki:
:raising_hand:

takahirom:
Assigned 👍

PR #432, opened 2020-01-19
Fix do not transition to SpeakerFragment when loading image is failure

n-seki:
## Issue
- related to #168

## Overview (Required)
- At https://github.com/DroidKaigi/conference-app-2020/pull/403, I call startPostponedEnterTransition when loading image is succeed.
- But I should call this method when loading image is failure. Otherwise we can not move to SpeakerFragment when (for example) network is off.
- Sorry I didn't notice 🙏 

## Links
- None

## Screenshot
- None


takahirom:
Thanks! LGTM!

takahirom:
Good catch!

PR #433, opened 2020-01-19
Use language change livedata for language change seamless load

### LINKED ISSUE #124: The language of the announcement should also change dynamically when switching languages
## Kind (Required)
Improvement

 ## Overview (Required)
I think that the language of the announcement should also change dynamically when switching devices language.

 ## Links	
none

chigichan24:
## Issue
- close #124 

## Overview (Required)
- ~Create `fun load()` in AnnoucementViewmodel.~
- ~Declare mutableLivedata for load function.~
- ~In `onViewCreated`, call `viewModel.load()` .~
- Create mutableLivedata that manage defaultLang.
- When `AnnoucementFragment#OnViewCreated()` called, update defaultLanguage.

## Links
- Nothing

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/announcement/src/main/java/io/github/droidkaigi/confsched2020/announcement/ui/viewmodel/AnnouncementViewModel.kt
+++ b/feature/announcement/src/main/java/io/github/droidkaigi/confsched2020/announcement/ui/viewmodel/AnnouncementViewModel.kt
@@ -3,17 +3,19 @@ package io.github.droidkaigi.confsched2020.announcement.ui.viewmodel
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
-import androidx.lifecycle.viewModelScope
+import androidx.lifecycle.asLiveData
+import androidx.lifecycle.distinctUntilChanged
+import androidx.lifecycle.liveData
+import androidx.lifecycle.switchMap
 import com.squareup.inject.assisted.AssistedInject
+import io.github.droidkaigi.confsched2020.model.repository.AnnouncementRepository
 import io.github.droidkaigi.confsched2020.ext.combine
 import io.github.droidkaigi.confsched2020.ext.toAppError
 import io.github.droidkaigi.confsched2020.ext.toLoadingState
 import io.github.droidkaigi.confsched2020.model.Announcement
 import io.github.droidkaigi.confsched2020.model.AppError
 import io.github.droidkaigi.confsched2020.model.LoadState
-import io.github.droidkaigi.confsched2020.model.repository.AnnouncementRepository
-import kotlinx.coroutines.flow.collect
-import kotlinx.coroutines.launch
+import io.github.droidkaigi.confsched2020.model.defaultLang
 import timber.log.Timber
 import timber.log.debug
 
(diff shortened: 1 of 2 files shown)


takahirom:
👀

feature/announcement/src/main/java/io/github/droidkaigi/confsched2020/announcement/ui/viewmodel/AnnouncementViewModel.kt
```diff
@@ -57,6 +50,22 @@ class AnnouncementViewModel @AssistedInject constructor(
         )
     }
 
+    fun load() {
+        viewModelScope.launch {
+            try {
+                announcementRepository.refresh()
```
takahirom: Thanks! 😉 
When you open it for the second time or later, you have to wait to read the acquisition from the API.
I think that Fragment is regenerated by ConfigrationChange, but I'm worried why the language is not adapted.
I think View could solve this problem. 👀

chigichan24: > I think View(Inside Fragment) could solve this problem. 👀

What do you mean?

takahirom: I'm thinking if the language can be switched by redisplaying in View instead of in ViewModel. 👀

takahirom: Please let me know if you have any problems doing it. 🙏

chigichan24: I understand. but I have no idea to redisplay only using View(Inside Fragment)😭 I'm going to think another solution.

chigichan24:
@takahirom  I checked the ViewModel lifecycle. When changed language configuration outside of the app, `AnnouncementViewModel#onCleared()`, is not called.  
It means that `AnnoucemntFragment`'s view is regenerated, but the ViewModel is not regenerated. The view refers to an old language setting value.  So I think it has to load the newest value expressly. What do you think?

takahirom:
I thought there was data in LocaledString. 🙇

takahirom:
But the API is different.

takahirom:
What about this way?

```kotlin
    private val languageLiveData = MutableLiveData(defaultLang())

    private val announcementLoadStateLiveData: LiveData<LoadState<List<Announcement>>> = languageLiveData.switchMap {lang->
        liveData<LoadState<List<Announcement>>> {
            emitSource(
                announcementRepository.announcements(lang)
                    .toLoadingState()
                    .asLiveData()
            )
            try {
                announcementRepository.refresh()
            } catch (e: Exception) {
                // We can show announcements with cache
                Timber.debug(e) { "Fail announcementRepository.refresh()" }
            }
        }
    }
```

chigichan24:
It's a nice way!! I'm going to change source code :muscle:

PR #447, opened 2020-01-19
Create iOS Sessions View and Initialize Project

### LINKED ISSUE #134: iOS App session list
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)

- Probably you can access sessions now by Kotlin Native
https://github.com/DroidKaigi/conference-app-2020/blob/master/ios-base/DroidKaigi%202020/ViewController.swift#L17
- You can introduce RxSwift etc. You can design iOS App Architecture 😄 

 ## Links	

-

ry-itto:
## Issue
- close #134 

### Initialize iOS Project
  - Use CocoaPods for library management
    - Use RxSwift and MaterialComponents
  - Use XcodeGen to generate `xcodeproj` file.
    - I make `Makefile` to execute `xcodegen` command and `bundle exec pod install` as alias.
### About App Architecture
  - Use MVVM
  - This PR implements only Session(Timeline) view.
  - Not using `Storyboard`, but using `xib` s
#### Directory structure
```
DroidKaigi 2020
├── AppDelegate.swift
├── ApplicationScheme.swift
├── Assets.xcassets
├── Common
├── Extentions
├── Info.plist
├── SceneDelegate.swift
└── Session // Feature directory
    ├── Models
    ├── ViewModels
    └── Views
```

Files are placed at feature directories. 

### Others
- Not support Dark mode yet.

## Links
- [Material Design -iOS](https://material.io/develop/ios/)
- [Material Design icons](https://github.com/google/material-design-icons)
  - use icons from here
- [Swift：UIColorを16進数カラーコードで初期化する](https://qiita.com/Kyome/items/eae6216b13c651254f64)

## Screenshot
|Sessions View (Expanded)|Sessions View (Contracted)| MyList (Empty) |
|:--: | :--:|:--:|
|Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-19 at 21 41 17|Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-19 at 21 41 29|Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-19 at 21 41 20|


README.md
```diff
@@ -232,5 +232,6 @@ This project uses some modern Android libraries and source codes.
   * [AutoClearedValue](https://github.com/android/architecture-components-samples/blob/9826b59956eb93c9e627bdf16a19a1c8bc28ce14/GithubBrowserSample/app/src/main/java/com/android/example/github/util/AutoClearedValue.kt)
 
 ### iOS
-
-// TODO: Add iOS Libraries
+* [XcodeGen](https://github.com/yonaskolb/XcodeGen)
+* [material-components-ios](https://github.com/material-components/material-components-ios)
+* [RxSwift](https://github.com/ReactiveX/RxSwift)
```
takahirom: Great 🥇

ios-base/DroidKaigi 2020/ApplicationScheme.swift
```diff
@@ -0,0 +1,55 @@
+//
+//  ApplicationScheme.swift
+//  DroidKaigi 2020
+//
+//  Created by 伊藤凌也 on 2020/01/17.
```
takahirom: I'm very grateful, but this repository does not currently include the author's name in source code comments. 🙇 
But If it is difficult to remove, this is fine.

ry-itto: ok. I will do! 👍 
change files are not so much, so it's not burdensome!

ios-base/DroidKaigi 2020/Session/Models/Session.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  Session.swift
+//  DroidKaigi 2020
+//
+//  Created by 伊藤凌也 on 2020/01/18.
+//
+
+import api
+
+struct Session {
```
takahirom: So sorry. There is ResponseToModelMapper.kt and probably you can use it 🙇 
So you can use Model class🙇 
Can you try it 🙏

takahirom: I think it was really hard. So sorry 🙇

ios-base/build_kotlin_modules.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cd ../
+./gradlew :data:api:linkIosX64 -PXCODE_CONFIGURATION=${CONFIGURATION}
```
takahirom: Probably the build phase added. 👀 
https://github.com/DroidKaigi/conference-app-2020/pull/400

takahirom:
Now we can not use ResponseToModelMapper. Because of that class in repository module. I forgot about it 🙇 
I will investigate this issue.

takahirom:
I was able to convert the response to a model.  Can you use this? 👀 
https://github.com/DroidKaigi/conference-app-2020/commit/d32222ab92f4f53ba4aa360d4f6d2455f8f3b3ef

ry-itto:
I will try this. 👍 
https://github.com/DroidKaigi/conference-app-2020/commit/d32222ab92f4f53ba4aa360d4f6d2455f8f3b3ef

ios-base/build_kotlin_modules.sh
ry-itto: Actually, I deleted `project.pbxproj` file that is generated by XcodeGen. So I added this code.

takahirom: Thanks! Since it is possible to change Kotlin files during the development of DroidKaigi's iOS app, it may be better to write the module build in the build phase. (If you build once, you can build faster with cache)
However, this is left to you.

ry-itto:
Can I call API from iOS project now?
I can't find... 😢

takahirom:
👀

takahirom:
Thanks! I fixed it 🙏 
https://github.com/DroidKaigi/conference-app-2020/commit/c93f8e486d80603233674d8ca3384b7369f24f8e

ry-itto:
Thank you! It works!

takahirom:
I fixed a little 🙇 (delete unneeded println)
https://github.com/DroidKaigi/conference-app-2020/commit/e4f580c6bfacaa05392611eae605451c1eacb5cc

ry-itto:
I think there are no way to get session's speakers from ios_combine.Session class. 😢

takahirom:
Thanks! I think you can use it like this.👀
https://github.com/DroidKaigi/conference-app-2019/blob/234e76c8961f5f6036cb5c248bbddfb67e63cd7e/frontend/ios/DroidKaigi%202019/Scenes/Session/SessionTableViewCell.swift#L29

ry-itto:
Thanks! I could use it and fixed. 👍

ry-itto:
I found a warning about AutoLayout, so I fixed now 🙇

ios-base/Podfile
```diff
@@ -0,0 +1,22 @@
+# Uncomment the next line to define a global platform for your project
+platform :ios, '13.2'
+
+target 'DroidKaigi 2020' do
+  # Comment the next line if you don't want to use dynamic frameworks
+  use_frameworks!
+  pod 'MaterialComponents'
+  pod 'RxSwift'
```
417-72KI: Why don't you use RxSwift via SwiftPM instead of CocoaPods?
XcodeGen supports SwiftPM since 2.8.0

ios-base/DroidKaigi 2020/Session/Models/SessionDataProvider.swift
```diff
@@ -0,0 +1,16 @@
+import ios_combined
+import RxSwift
+
+final class SessionDataProvider {
+    func fetchSessions() -> Observable<[Session]> {
+        return Observable.create { observer -> Disposable in
+            ApiComponentKt.generateDroidKaigiApi().getSessions(callback: { response in
+                let model = ResponseToModelMapperKt.toModel(response)
+                observer.onNext(model.sessions)
```
takumaosada: Why don't you use `Single` instead of `Observable` because of just a HTTP get response.
Maybe, just add `onComplete` after `onNext` might be fine.

takahirom:
@takumaosada Thank you for your review!!👍
"Why don't" is a bit of a strong expression, so "it seems better" that this might be better🙇🙇🙏🙏

ios-base/DroidKaigi 2020/Session/ViewModels/SessionViewModel.swift
```diff
@@ -0,0 +1,47 @@
...
+import RxSwift
+import RxCocoa
+
+final class SessionViewModel {
+
+    private let disposeBag = DisposeBag()
+
+    // input
+    let viewDidLoad = PublishRelay<Void>()
+    let toggleEmbddedView = PublishRelay<Void>()
```
417-72KI: I recommend to make `~ Relay` as `private` to prevent unexpected events or subscribing.
If you want to use it from outside, you may create wrapper property or function like this.

```swift
private let viewDidLoadRelay = PublishRelay<Void>()

// to fire event
func viewDidLoad() {
    viewDidLoadRelay.accept(())
}

// to subscribe
var viewDidLoad: Observable<Void> {
    viewDidLoadRelay.asObservable
}
```

ios-base/Podfile
ry-itto: I thought it wouldn't be nice to distribute library dependencies. So I use CocoaPods only.

ios-base/DroidKaigi 2020/Session/Models/SessionDataProvider.swift
ry-itto: I think so too. I will change the code to use `Single`. Thanks! 👍

ios-base/DroidKaigi 2020/Session/ViewModels/SessionViewModel.swift
ry-itto: Certainly, I think this looks better than the current one. I will fix it. 
Thanks! 👍

ios-base/Makefile
```diff
@@ -0,0 +1,5 @@
+init:
+	xcodegen
+	bundle exec pod install
```
YusukeHosonuma: The command `bundle install (--path vendor/bundle)` is missing? (or execute in manually?)

ry-itto: Sorry, I missed. I will add `bundle install --path vender/bundle` 🙏.
Thank you for reporting! 😄

ios-base/Gemfile.lock
```diff
@@ -0,0 +1,82 @@
...
+      nanaimo (~> 0.2.6)
+
+PLATFORMS
+  ruby
+
+DEPENDENCIES
+  cocoapods
+
+BUNDLED WITH
+   2.1.1
```
jmatsu: It would be nicer if bundler version is the same with `.ci/Gemfile.lock` or should be at least v1 to unify the build environment. bundler 2.x is not incompatible with 1.x without bundler configs.

ios-base/Gemfile
```diff
@@ -0,0 +1 @@
+gem 'cocoapods'
```
YusukeHosonuma: `source 'https://rubygems.org'` are missing?

I got a bellow error:

```
Your Gemfile has no gem server sources. If you need gems that are not already on your machine, add a line like this to your Gemfile:
source 'https://rubygems.org'
```

jmatsu: Can you please add *gem source* and ruby version? You can copy it from `.ci/Gemfile`.

ry-itto: I'm sorry, It's missed too. I will add it.

ios-base/Makefile
jmatsu: Good catch. It would be nicer if `bundle install --path vendor/bundle` is also executed by this script.

ios-base/Makefile
```diff
@@ -0,0 +1,5 @@
+init:
```
jmatsu: I would like you to write a getting started for iOS section in README.md.

In addition, is this file for GNU make? If so, please note it as a required dependency.

ios-base/build_kotlin_modules.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cd ../
```
jmatsu: `cd "$(git rev-parse --show-toplevel)"` sounds better. It allows moving up to the repository root directory.

ios-base/build_kotlin_modules.sh
```diff
@@ -0,0 +1,4 @@
...
+./gradlew :ios-combined:linkIosX64 -PXCODE_CONFIGURATION=${CONFIGURATION}
```
jmatsu: XCODE_CONFIGURATION looks better than CONFIGURATION

jmatsu: I'm just curious about how many times we should execute this script. Is running once enough basically? 👀

ios-base/Gemfile.lock
ry-itto: OK. I did it! 👍

ios-base/Gemfile
ry-itto: I did it👍

ios-base/build_kotlin_modules.sh
ry-itto: > XCODE_CONFIGURATION looks better than CONFIGURATION

I'm sorry, it doesn't work. 🙏 I will set configuration by variable `XCODE_CONFIGURATION `


> I'm just curious about how many times we should execute this script. Is running once enough basically? 👀

I think this script should execute once enough basically. When Kotlin module is changed, we should execute this script.

ios-base/Makefile
ry-itto: > I would like you to write a getting started for iOS section in README.md.

In this case, README.md represents Project's README.md ? not ios-bases/README.md?

takahirom: I think it seems better to use `ios-base/README.md`. 
Because Android Engineers will do `make init` without thinking 😄 
But it is good to link `ios-base/README.md` from `root/README.md`

ios-base/Makefile
```diff
@@ -1,5 +1,6 @@
 init:
 	xcodegen
```
417-72KI: [q]
How is XcodeGen expected to install?(Mint or HomeBrew?)

It would be better using Mint, and create Mintfile.

ios-base/Makefile
ry-itto: I will write `README.md` at `ios-base/README.md` .
I think it is natural. 🤔

ios-base/Makefile
ry-itto: I don' t think XcodeGen's install way. 😢 (I installed by HomeBrew)

Why is using Mint better?

417-72KI: Mint (especially using Mintfile) can fix the version of XcodeGen like dependency tools(e.g. CocoaPods, Carthage) 
Considering the release frequency of XcodeGen, I think it's version is better to be controlled.

ios-base/Podfile
417-72KI: I'm sorry but I couldn't understand your intention. 
What does 'distribute library dependencies' mean?

ios-base/build_kotlin_modules.sh
417-72KI: Build phase can be set in `project.yml` with `prebuildScripts` key ([ref](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md#build-script)).
Is that useless?

ios-base/Podfile
ry-itto: I'm sorry I couldn't tell my opinion correctly...

In Japanese, 
ライブラリの依存関係を分散させることがあまり良くないことだと考えており、そのことから、今回はCocoaPodsのみを使用しています。

ios-base/Makefile
ry-itto: I see. I will add Mintfile to control XcodeGen's version. Thanks!

ios-base/Podfile
ry-itto: I'm sorry I couldn't tell my opinion correctly...

In Japanese,
ライブラリの依存関係を分散させることがあまり良くないことだと考えており、そのことから、今回はCocoaPodsのみを使用しています。

ios-base/build_kotlin_modules.sh
ry-itto: I think it is useful.
I will try execute this script Xcode Project's prebuild phase instead of Make.

ios-base/DroidKaigi 2020/Session/Views/SessionPageViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+    case day2 = 2
+    case myPlan = 3
+
+    var date: Date? {
+        let calendar = Calendar(identifier: .gregorian)
+        switch self {
+        case .day1:
+            return calendar.date(from: .init(year: 2020, month: 2, day: 20))
+        case .day2:
+            return calendar.date(from: .init(year: 2020, month: 2, day: 20))
```
woxtu:
```suggestion
            return calendar.date(from: .init(year: 2020, month: 2, day: 21))
```

ry-itto: Oh... Thank you!
I will take this suggested changes.

ios-base/DroidKaigi 2020/Session/Views/SessionViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+            let templateFilterListImage = filterListImage?.withRenderingMode(.alwaysTemplate)
+            filterButton.setImage(templateFilterListImage, for: .selected)
+            let arrowUpImage = UIImage(named: "ic_keyboard_arrow_up")
+            let templateArrowUpImage = arrowUpImage?.withRenderingMode(.alwaysTemplate)
+            filterButton.setImage(templateArrowUpImage, for: .normal)
+            filterButton.setTitle("Filter", for: .selected)
+            filterButton.setTitle("", for: .normal)
+        }
+    }
+    @IBOutlet weak var collectionView: UICollectionView! {
```
woxtu: Just out of my curiosity, why did you use `UICollectionView` instead of `UITableView`?

ry-itto: At the time I implemented this, I would use [MDCBaseCell](https://material.io/develop/ios/components/list/api-docs/Classes/MDCBaseCell.html) or [MDCSelfSizingStereoCell](https://material.io/develop/ios/components/list/api-docs/Classes/MDCSelfSizingStereoCell.html). This is inherit class of `UICollectionViewCell`. So here uses `UICollectionView` .

ios-base/build_kotlin_modules.sh
ry-itto: I fixed it on https://github.com/DroidKaigi/conference-app-2020/pull/447/commits/2111bd7f13b0ac2963a14c516488dd49aeacc75d ! Please check it. 🙏

ios-base/build_kotlin_modules.sh
ry-itto: This is no longer needed by https://github.com/DroidKaigi/conference-app-2020/pull/447/commits/2111bd7f13b0ac2963a14c516488dd49aeacc75d 🙏

ios-base/build_kotlin_modules.sh
ry-itto: This is no longer needed by https://github.com/DroidKaigi/conference-app-2020/pull/447/commits/2111bd7f13b0ac2963a14c516488dd49aeacc75d 🙏

ios-base/Makefile
jmatsu: Agreed. It doesn't sound that we need to put the section I've mentioned into `/path/to/repo/README.md`. `/path/to/repo/ios-base/README.md` is fine.

ios-base/Makefile
ry-itto: I did it 👍 https://github.com/DroidKaigi/conference-app-2020/pull/447/commits/fa63299e638a4a611f47cd94262417f83e81e172

ios-base/DroidKaigi 2020/Session/Views/SessionViewController.swift
woxtu: I completely understand. Thanks.

ios-base/build_kotlin_modules.sh
jmatsu: I guess modifying Kotlin modules will happen especially in the early stage. To avoid some troubles, I suppose it would be nicer if this script would be executed in the build phases like cocoapods' linking. What do you think about it? I'm just an iOS newbie so I can only say from my imagination though...

ry-itto:
@woxtu 
I fixed 👍

ios-base/build_kotlin_modules.sh
jmatsu: ah, make sense. cool.

ios-settings.gradle
```diff
@@ -0,0 +1 @@
+include ':ios-combined'
```
jmatsu: Is this required by default?  ref: https://github.com/DroidKaigi/conference-app-2020/pull/447/files#diff-900def4b7f2013a4665d633b21528b04R3

ios-base/build_kotlin_modules.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+echo "include ':ios-combined'" > $SRCROOT/../ios-settings.gradle
```
jmatsu: If you would like to include this project only during ios build, there is another solution. At first, please add the following to settings.gradle.

```
if (System.getenv("BUILD_IOS") == "true") {
  include ':ios-combined'
}
```

and then, run `BUILD_IOS=true ./gradlew ios-combined:xxx`. ios-combined would be resolved only while BUILD_IOS=true was supplied.

ios-settings.gradle
takahirom: I added this file. If Android Studio loads this. The duplicate file error will be occurring 😭

ios-base/build_kotlin_modules.sh
takahirom: 🙇

ry-itto: Is this solution for only represents
```
echo "include ':ios-combined'" > $SRCROOT/../ios-settings.gradle
```
line? Does this solution mean the line can be removed?

jmatsu: Yeah, exactly.

ry-itto: I see. OK. I will use that solution.

ry-itto: I've done above. 👍 https://github.com/DroidKaigi/conference-app-2020/pull/447/commits/251f9da59d936c57a3ad464b6b8490e68b586579

koogawa:
> Not support Dark mode yet.

OK 👌 We should set `UIUserInterfaceStyle` to `Light` in our Info.Plist until support Dark mode. What do you think? 😃

koogawa:
If we support Landscape mode, need to check a lot of things! 



e.g. Lef and Right margin of SessionCell

Simulator Screen Shot - iPhone 8 - 2020-01-21 at 00 42 29

I think That's no problem 😄  Because Android app also supports landscape mode.

ry-itto:
@koogawa 
Thank you for reporting! 😄 

First, 
> > Not support Dark mode yet.
> 
> OK 👌 We should set `UIUserInterfaceStyle` to `Light` in our Info.Plist until support Dark mode. What do you think? 😃

I think I should set `UIUserInterfaceStyle` to `Light`. So I will do it. 👍 

And about landscape mode.

> If we support Landscape mode, need to check a lot of things!

Certainly Android App support landscape mode, but I think we should disable landscape mode.
Reasons are there.
1. Need to check a lot of things.
2. In this App, landscape mode is not suitable. (I think)

So I want to disable landscape mode. What do you think?

ios-base/Makefile
ry-itto: I added README.md and update Project's README.md! Please check it. 🙏 
https://github.com/DroidKaigi/conference-app-2020/pull/447/commits/c465d8c6b23f23c75e9317236ddf56f9dfe73ce0

koogawa:
@ry-itto Yeah, I agree👍😃

takahirom:
Can you merge from master?🙏

ios-base/Podfile
417-72KI: I see.
I think it is bothered that MaterialComponents only supports CocoaPods😅

ios-base/build_kotlin_modules.sh
417-72KI: LGTM!

ios-base/Podfile
ry-itto: Yes, I think so too. :tired_face:

ry-itto:
@koogawa 
I fixed. 👍Please check it 🙏 

@takahirom 
I merged master into this branch. Please check it 🙏

takahirom:
@ry-itto Can you create an issue for some of the features you might need? 🙏

ry-itto:
@takahirom 
Yes, of course I can. 👍 
I will create some issues.

PR #449, opened 2020-01-19
Show firebase topic notification

### LINKED ISSUE #145: Show firebase topic notification
## Kind (Required)

- Improvement

 ## Overview (Required)

- Show firebase topic notification 🔥
- Probably we can use implementation of 2019

 ## Links	

- https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/notification/src/main/java/io/github/droidkaigi/confsched2019/notification/worker/ManageTopicSubscriptionWorker.kt

entaku19890818:
## Issue
- close #145

## Overview (Required)
- Show firebase topic notification

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/res/layout/fragment_bottom_sheet_sessions.xml
+++ b/feature/session/src/main/res/layout/fragment_bottom_sheet_sessions.xml
@@ -11,6 +11,7 @@
         android:layout_height="match_parent"
         tools:context="io.github.droidkaigi.confsched2020.session.ui.SessionPagesFragment"
         >
+
         <androidx.constraintlayout.widget.Guideline
             android:id="@+id/guideline_start"
             android:layout_width="wrap_content"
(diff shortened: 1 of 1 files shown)


android-base/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Sun Jan 19 20:22:02 JST 2020
```
takahirom: Thank you! 👍 
This code is for the root directory. So we don't need this file. 🙇

android-base/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
takahirom: This code is for the root directory. So we don't need this file. 🙇

android-base/src/main/java/io/github/droidkaigi/confsched2020/App.kt
```diff
@@ -7,13 +7,21 @@ import com.google.firebase.firestore.FirebaseFirestore
...
+import io.github.droidkaigi.confsched2020.ext.changedForever
+import io.github.droidkaigi.confsched2020.model.SystemProperty
+import io.github.droidkaigi.confsched2020.system.store.SystemStore
 import timber.log.LogcatTree
 import timber.log.Timber
+import javax.inject.Inject
 
 open class App : DaggerApplication(), AppComponentHolder {
+    @Inject
+    lateinit var systemStore: SystemStore
```
takahirom: DroidKaigi 2020 does not use Flux. So we should not use Store and Actions.
This fix will likely be complicated.
I mistakenly attached the easy tag to this issue. 🙇 
It's my mistake, so I think you can stop this PR.

entaku19890818: Should I do nothing?

android-base/gradle/wrapper/gradle-wrapper.properties
entaku19890818: delete it

android-base/gradlew
entaku19890818: Sorry. I can't checked.
delete it

android-base/src/main/java/io/github/droidkaigi/confsched2020/App.kt
takahirom: There are many things to erase.  Since the architecture is not Flux and it is not possible to change the language from the settings at first, it is necessary to remove Action, SystemStore, Dispatcher, CoroutinePlugin, unused LiveData extension functions and functions that depend on them.  Since the amount of change required is large, I thought about closing this PR once.  But if you just delete it, you may not need to make such a big change.

takahirom:
@entaku19890818  Please let me know if you have any problems. 🙇

entaku19890818:
Thank you for your support!
@takahirom


I do not understand the following meaning.  Need to turn off the process?

 > But if you just delete it, you may not need to make such a big change.

entaku19890818:
Does it mean that only the following files are changed?

https://github.com/entaku19890818/conference-app-2020/blob/0c4d7ae3f214b59fa10bffbe193af934f8db405d/feature/announcement/src/main/java/io/github/droidkaigi/confsched2020/announcement/ui/AnnouncementTopic.kt

takahirom:
@entaku19890818 
Sorry for Japanese 🙇 
日本語で失礼します。
このプルリクエストには不要と思われるファイルが現状たくさん入っています。
これを削除してなくしていくことが必要となります。
例えば最初は設定を保存しないためSystemStoreやSystemActionは必要ありません。またそれを使うために必要なったクラスたち(Action, SystemStore, Dispatcher, CoroutinePlugin, unused LiveData extension functions)も消す必要があります。
ただ消せばそれなりに形にはなるので、お願いできますか？ 🙏 
もし難しければ言ってくだされば大丈夫です 🙇 
よろしくおねがいします。

entaku19890818:
はいいったん消して見ます！

entaku19890818:
新しいプルリクで削除したものを出しました。
notification　部分のみの追加になっていると思います。確認お願いします
@takahirom

PR #451, opened 2020-01-19
Scroll to current time session when tapping tab

### LINKED ISSUE #152: Scroll to current time session when tapping tab
## Kind (Required)

- Improvement

 ## Overview (Required)

- Probably We can use tab reselect events. 📜

tetsukay:
## Issue

- close #152 

## Overview (Required)

- Scroll to the current session, when tapping tab.

## Screenshot
Before | After(After tapping a tab)
:--: | :--:
 |


takahirom:
Thank you!

> Scroll to the current session, when session list is created.
Do not scroll by tapping a tab. I could not feel a strong need.
Is it necessary?

Sorry, There is another issue that moves to the time of the current session when opened. https://github.com/DroidKaigi/conference-app-2020/issues/150
For this reason, I was thinking about doing an issue where the tab would scroll when pressed.

tetsukay:
Thanks review, I understand.
I will additional commits.

tetsukay:
@takahirom 
Hi, I changed to tap tab and scroll to current session.
Please I changed to tap tab and scroll to current session., 🙏

PR #452, opened 2020-01-19
Change highlight color to white

### LINKED ISSUE #111: Add menu icon tap feedback
## Kind (Required)

- Improvement

 ## Overview (Required)

- Add menu icon tap feedback
- Change highlight to the same color as the tab highlight

 ## Links	

-

kenjikitamura:
## Issue
- close #111

## Overview (Required)
- Change highligh color to white(tab highlight color)

## Screenshot
Before | After
:--: | :--:
 |


matsujun:
👀

matsujun:
Currently, the background and text color of Toolbar is decided based on `isIndigoBackground` .
If `isIndigoBackground` is true, the background color is indigo (colorPrimarySurface). 
It's OK to use white (colorOnPrimarySurface).
But if `isIndigoBackground` is false, the background color is white (colorSurface). 😢 
In this case, we'd better to use other color for highlight. (e.g. same as `app:titleTextColor` .)

kenjikitamura:
Change highlight color to default highlight color when isIndigoBackground is false.

Navigation button
スクリーンショット 2020-01-26 23 16 04

Search close button
スクリーンショット 2020-01-26 23 16 32

android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
```diff
@@ -172,6 +179,30 @@ class MainActivity : DaggerAppCompatActivity() {
...
+            }
+        }
+        return super.onCreateOptionsMenu(menu)
+    }
+
+    private fun setRippleColor(view: View, isIndigoBackground: Boolean) {
+        (view.background as? RippleDrawable)?.setColor(
+            ColorStateList.valueOf(this.getThemeColor(
+                if (isIndigoBackground) {
+                    R.attr.colorSurface
```
matsujun: If we use DarkTheme, colorSurface is invisible when isIndigoBackground is true.
colorPrimarySurface looks better.

takahirom:
👀

android-base/src/main/java/io/github/droidkaigi/confsched2020/MainActivity.kt
takahirom: Can you use `colorPrimarySurface`??

kenjikitamura: I think colorPrimarySurface cannot be seen because it is the same color as the background.
screenshot

I think using white color is fine.
white

DarkTheme
white2

What are your thoughts on that?

matsujun: The color of icons seems to changed from `colorPrimarySurface` to `colorOnPrimary` on base code.
https://github.com/DroidKaigi/conference-app-2020/pull/452/files#diff-14153ff2fbf8b01642f55ee0af1ccdc7L219-L225

And the definition of `colorOnPrimary` is white on both light and dark theme.
https://github.com/DroidKaigi/conference-app-2020/search?q=colorOnPrimary&unscoped_q=colorOnPrimary

So I agree, Its good to use white color for ripple. It's same as icon color.
But I prefer to use semantic color definition (likes R.attr.colorOnPrimary etc) instead of direct color definition (Color.WHITE) to control colors by theme.

kenjikitamura: Use colorOnPrimary for ripple color.

70dcb376f15c2ceb54064b2df512830a5f479b13

PR #455, opened 2020-01-19
Use target sdk version 29 and compile sdk version 29

### LINKED ISSUE #454: Use targetSdkVersion 29 and compileSdkVersion 29
## Kind (Required)

- Improvement

 ## Overview (Required)

- There is no reason why it is currently API Level 28, and I just forgot it. 😄 

 ## Links	

-

tatsuya-okayama:
## Issue
- close #454 

## Overview (Required)
- change targetSdkVersion and compileSdkVersion to 29
- Because BuildCompat.isAtLeastQ became Deprecated, use Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q.
- ~change val to const val as in Packages.kt~

I tried to fix the two below, but I will revert if there is a problem !! 🙇

### DIFF
--- a/buildSrc/src/main/java/dependencies/Versions.kt
+++ b/buildSrc/src/main/java/dependencies/Versions.kt
@@ -1,8 +1,8 @@
 package dependencies
 
 private object Versions {
-    val androidTargetSdkVersion = 28
-    val androidCompileSdkVersion = 28
+    val androidTargetSdkVersion = 29
+    val androidCompileSdkVersion = 29
     val androidMinSdkVersion = 21
 
     private val versionMajor = 1
--- a/buildSrc/src/main/java/dependencies/Versions.kt
+++ b/buildSrc/src/main/java/dependencies/Versions.kt
@@ -1,16 +1,16 @@
 package dependencies
 
 private object Versions {
-    val androidTargetSdkVersion = 29
-    val androidCompileSdkVersion = 29
-    val androidMinSdkVersion = 21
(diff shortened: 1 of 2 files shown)


takahirom:
👀
https://developer.android.com/about/versions/10/behavior-changes-10

buildSrc/src/main/java/dependencies/Versions.kt
```diff
@@ -1,16 +1,16 @@
 package dependencies
 
 private object Versions {
-    val androidTargetSdkVersion = 28
-    val androidCompileSdkVersion = 28
-    val androidMinSdkVersion = 21
+    const val androidTargetSdkVersion = 29
+    const val androidCompileSdkVersion = 29
+    const val androidMinSdkVersion = 21
```
takahirom: I understand both opinions about cost.  👀
For example, Dep.kt has a problem that cost cannot be used because it refers to other variables.  Also, to keep many dependencies simple, we operate with no cost.  
Please let me know if you have any opinions🙏

tatsuya-okayama: > For example, Dep.kt has a problem that cost cannot be used because it refers to other variables. 

→ Can all vals of Dep.kt be changed to const val ? ( except for hyperionPlugins )

but... I overlooked using arrays...
This change will revert 🙇

takahirom:
Thanks! LGTM 👍

PR #456, opened 2020-01-19
Add floor map image

### LINKED ISSUE #43: Implement floor map screen without design
## Overview (Required)
Title says everything.

eyedol:
## Issue
- close #43 

## Overview (Required)
- Add PhotoView Library to offer zoom support
- Add jitpack repo as it's needed by the PhotoView library 
- Add floor image exported from figma as AVD optimized with [avocado](https://github.com/alexjlockwood/avocado) 🥑  tool

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


feature/floormap/src/main/res/layout/fragment_floormap.xml
```diff
@@ -11,6 +12,18 @@
         tools:context=".ui.FloorMapFragment"
         >
 
+        <com.github.chrisbanes.photoview.PhotoView
```
takahirom: 👍

eyedol:
@takahirom  I'm not so familiar with the figma tool so the image I exported isn't sharp.  The resolution is a bit off. Could the floor map image be exported for the different DPIs? so I can just include them in the project?

takahirom:
@eyedol Since the image has not been decided yet, I think that it is good to put it in this size for the time being. 👍

takahirom:
Thank you! I don't know the cause but I will try to merge.

eyedol:
Makes sense. I'll use the AVD for the current map image then and make the PR ready for review 🙇

eyedol:
@takahirom  ready for a review

takahirom:
Thank you. Can you remove the pin once? 🙏

takahirom:
Thanks!
LGTM🗺️

PR #485, opened 2020-01-20
fix a placeholder image is not shown when offline at the speaker screen.

### LINKED ISSUE #313: Should show placeholder when loading user image at speaker screen.
## Kind (Required)

- Improvement

 ## Overview (Required)

- User placeholder icon does not displayed if user is using in offline and weak network connection.
    - In the other screens, it seems to be displaying placeholder when loading.

kazkn:
## Issue
- close #313

## Overview (Required)
- The speaker screen already had a placeholder icon. but it not had set that placeholder ImageView when starting the image loading by Coil.

### Offline(Speaker image is not loaded)
Before | After
:--: | :--:
 | 
### Online(Speaker image is loaded)
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SpeakerDetailItem.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SpeakerDetailItem.kt
@@ -1,5 +1,6 @@
 package io.github.droidkaigi.confsched2020.session.ui.item
 
+import android.content.Context
 import android.text.method.LinkMovementMethod
 import androidx.core.content.ContextCompat
 import androidx.lifecycle.LifecycleOwner
@@ -15,6 +16,7 @@ import io.github.droidkaigi.confsched2020.item.EqualableContentsProvider
 import io.github.droidkaigi.confsched2020.model.Speaker
 import io.github.droidkaigi.confsched2020.session.R
 import io.github.droidkaigi.confsched2020.session.databinding.ItemSpeakerDetailBinding
+import io.github.droidkaigi.confsched2020.util.lazyWithParam
 import javax.inject.Named
 
 class SpeakerDetailItem @AssistedInject constructor(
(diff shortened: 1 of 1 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SpeakerDetailItem.kt
```diff
@@ -48,12 +36,26 @@ class SpeakerDetailItem @AssistedInject constructor(
 
         speaker.imageUrl ?: onImageLoadedCallback()
 
+        val context = viewBinding.root.context
+        val placeHolder = VectorDrawableCompat.create(
```
takahirom: Thanks!
Can you use lazyWithParam method like this :+1:

```kotlin
    private val placeholder by lazyWithParam<Context, VectorDrawableCompat> { context ->
        VectorDrawableCompat.create(
            context.resources,
            R.drawable.ic_person_outline_black_32dp,
            null
        )?.apply {
            setTint(
                ContextCompat.getColor(context, R.color.speaker_icon)
            )
        }
    }
```

kazkn: Thanks for your advice. I will try this code:smile:

kazkn:
I fixed an issue commented by takahirom. please re-review. thanks.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SpeakerDetailItem.kt
```diff
@@ -48,12 +50,17 @@ class SpeakerDetailItem @AssistedInject constructor(
...
+        val context = viewBinding.root.context
+
         Coil.load(context, speaker.imageUrl) {
             crossfade(true)
-            placeholder(placeHolder)
+            placeholder(placeholder.get(context))
             transformations(CircleCropTransformation())
             lifecycle(lifecycleOwnerLiveData.value)
             target(
+                onStart = {
```
takahirom: 📝

PR #486, opened 2020-01-20
115: Add Contributors screen

### LINKED ISSUE #115: Contributor list screen
## Overview (Required)

 - We should implement contributor list screen
 - I want this feature early for contributor motivation.
image

## Links	

-

mkano9:
## Issue
- Close #115 

## Overview (Required)
- Added Contributors screen

## Screenshot
 |


data/repository/src/main/java/io/github/droidkaigi/confsched2020/data/repository/internal/DataContributorRepository.kt
```diff
@@ -4,32 +4,39 @@ import io.github.droidkaigi.confsched2020.data.api.DroidKaigiApi
...
 
-    private suspend fun contributorList() = contributorDatabase
-        .contributorList()
-        .map { it.toContributor() }
+    private fun contributorList(): Flow<List<Contributor>> =
+        contributorDatabase
+            .contributorList()
+            .map {
+                it.mapIndexed { index, entity ->
+                    entity.toContributor(index + 1)
```
mkano9: I could have used `sort` value from API, but I felt it didn't matter for the purpose of displaying a non-zero index order.

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
```diff
@@ -1,11 +1,87 @@
...
 class ContributorsViewModel @Inject constructor(
     private val contributorRepository: ContributorRepository
 ) : ViewModel() {
-    // TODO Implement
+
+    /**
+     * [ViewState] has to be wrapped to workaround a compile error about type mismatch.
+     * [combine] doesn't seem to understand child classes.
+     */
+    data class UiModel(val viewState: ViewState) {
```
mkano9: I know this deviates a bit from the pattern used in this repo. The challenge I encountered is the following:

When there's no cache yet, the LiveData emits an empty contributors list and then makes an API call to fetch the list. I need to keep the loading indicator visible since an API call technically could take a noticeable amount of time depending on the user's network condition. 

Long story short, I first tried to follow the existing UiModel pattern, but then IMO things started to become harder to understand as I have to think about more cases/states of `contributorsLoadStateLiveData`. e.g. Would have to take more conditions in consideration to decide when `isLoading` should be true or false.

Having said that, if you still prefer the existing pattern with UiModel having `isLoading`, `error`, and `contributors`, I can try to get that to work properly 👍

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
```diff
@@ -1,11 +1,87 @@
...
+    val uiModel: LiveData<UiModel> = combine(
+        initialValue = UiModel.DEFAULT,
+        liveData1 = contributorsLoadStateLiveData
+    ) { _, loadState ->
+        when (loadState) {
+            is LoadState.Loading -> UiModel(ViewState.Loading)
+            is LoadState.Error -> {
+                val appError = loadState.getErrorIfExists()?.toAppError()
+                if (contributors.isNotEmpty()) {
+                    UiModel(ViewState.Loaded(contributors))
```
mkano9: I wanted to introduce a `FullError` state where the screen would display a "Retry" button. This would be for the users who just installed the app and visited the Contributors page for the first time(no cache yet) when they happen to have bad network connection.

If you are interested, here is what I tried:

`
fun onRetry() {
    reloadContributorsLiveData.value = LoadingState.Loading
    viewModelScope.launch() {
        try {
            contributorRepository.refresh()
            reloadContributorsLiveData.value = LoadingState.Loaded
        } catch (exception: Exception) {
            reloadContributorsLiveData.value = LoadingState.Error(exception)
        }
    }
}
`

`reloadContributorsLiveData.value` is added as `liveData2` to the `uiModel`'s `combine(...)`. When `refresh()` executes, it doesn't trigger another event in `contributorsLoadStateLiveData `'s `emitSource`. I guess it's not possible to trigger an event to `emitSource` from outside the `liveData` builder function?

takahirom:
👀

takahirom:
Now speaker_icon.xml for contributors. Can you use contributor_icon.xml

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/item/ContributorItem.kt
```diff
@@ -0,0 +1,47 @@
...
+    private val lifecycleOwnerLiveData: LiveData<LifecycleOwner>
+) : BindableItem<ItemContributorBinding>(contributor.id.toLong()) {
+
+    override fun getLayout(): Int = R.layout.item_contributor
+
+    override fun bind(viewBinding: ItemContributorBinding, position: Int) {
+        viewBinding.contributor = contributor
+
+        val context = viewBinding.contributorImage.context
+        val placeholder = SpeakerPlaceholderCreator.create(context)
```
takahirom: You can use `lazyWithParam` method like this 👍 

```
    private val placeholder by lazyWithParam<Context, VectorDrawableCompat> { context ->
        VectorDrawableCompat.create(
            context.resources,
            R.drawable.ic_person_outline_black_32dp,
            null
        )?.apply {
            setTint(
                ContextCompat.getColor(context, R.color.speaker_icon)
            )
        }
    }
```

data/db/src/main/java/io/github/droidkaigi/confsched2020/data/db/internal/CacheDatabase.kt
```diff
@@ -28,7 +28,7 @@ import io.github.droidkaigi.confsched2020.data.db.internal.entity.StaffEntityImp
         (StaffEntityImpl::class),
         (ContributorEntityImpl::class)
     ],
-    version = 16
+    version = 17
```
takahirom: 🙏

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
```diff
@@ -1,11 +1,87 @@
...
+     * [ViewState] has to be wrapped to workaround a compile error about type mismatch.
+     * [combine] doesn't seem to understand child classes.
+     */
+    data class UiModel(val viewState: ViewState) {
+        companion object {
+            val DEFAULT = UiModel(ViewState.Loading)
+        }
+    }
+
+    sealed class ViewState {
```
takahirom: Thanks for the various challenges. 🏆 
Explain why I don't provide UiModel in such a shield class.
This is because the actual application needs to be very flexible, such as loading while displaying the cache. Such as display errors, but also display contents.
For that reason, I'm a little nervous about returning in a shielded class. Please let us know what you have.

takahirom: Also, with the current method using ContributorsViewModel.ViewState, if an error occurs from Loaded, the content will be displayed even though there is no data in UiModel. The state of the UiModel and the display state may not match.

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
takahirom: sorry. I did not read this. I'll check it. 🙏

takahirom: Thanks for checking in in various ways. 👍 
> Having said that, if you still prefer the existing pattern with UiModel having isLoading, error, and contributors, I can try to get that to work properly 👍

There may be some difficulties, but we want to check the validity of UiModel.

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
takahirom: 👀

takahirom: It would be better to be able to retry!
I think we should add `reloadContributorsLiveData` as you commented. 👀 
I have an implementation image like this.

In fragment

```
viewModel.uiModel.observe { uiModel ->
  if(uiModel.hasError) {
    showSnackbar(uiModel.error.message) { viewMode.retry() }
  }
}
```

mkano9:
> Thank you!
> Now speaker_icon.xml for contributors. Can you use contributor_icon.xml

Ops you're right, contributor is not a speaker :)

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
mkano9: Sounds good. I will look into it in a separate PR.

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/item/ContributorItem.kt
mkano9: Oh this is cool 😎

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
mkano9: Thanks for the feedback. I will follow the existing pattern used on other screens.

Re: 
>Also, with the current method using ContributorsViewModel.ViewState, if an error occurs from Loaded, the content will be displayed even though there is no data in UiModel. The state of the UiModel and the display state may not match.

I think the state of the UiModel and what's displayed matches in my approach. The idea is that view renders whatever state the uiModel LiveData emits. A state represents a possible state the view can render independent from what was emitted previously. For example, emitting both an error and `isLoading=true` is supposed to be impossible.

If an error occurs and the ViewModel already has the contributors list, it will emit the list. (or can make it emit the list together with `AppError`, but other screens don't seem to do that) If it has never been able to fetch the contributors list, it will emit the error.

mkano9:
@takahirom Thanks for your feedback :) It's ready for review again.

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
```diff
@@ -1,11 +1,60 @@
...
+        emitSource(
+            contributorRepository.contributorContents()
+                .toLoadingState()
+                .asLiveData()
+        )
+
+        try {
+            contributorRepository.refresh()
+        } catch (exception: Exception) {
+            // We can show contributors with cache
```
mkano9: This approach has a problem and I believe the same issue exists on other screens.

This is the sequence `contributorsLoadStateLiveData` emits when the screen opens for the first time with no previous cache.
UiModel(isLoading=true, error=null, contributors=[])
UiModel(isLoading=false, error=null, contributors=[])
UiModel(isLoading=false, error=null, contributors=[Contributor(id=1386930, name=takahirom, iconUrl=https://github.com/takahirom.png, profileUrl=https://github.com/takahirom, rankOrder=1) ...])

The second one dismisses the loading indicator because the repository emits an empty list when no cache exists. When the API call takes longer, there is a state where the screen shows nothing until the API call successfully completes. 

In my previous commits, I was emitting an error LiveData right here. I noticed that it causes another issue. When I have cache in the DB and open the screen with no network(or any errors fetching API response), the screen intermittently does not load data from cache. I'm guessing it's caused by some kinda racing condition in `emitSource`...?

I left it as is for now so that it can be tackled as a separate PR since it's an existing issue on other screens.

takahirom:
👀

feature/contributor/src/main/res/layout/item_contributor.xml
```diff
@@ -0,0 +1,66 @@
...
+        >
+
+        <TextView
+            android:id="@+id/contributor_offset_text"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:minWidth="36dp"
+            android:gravity="center"
+            android:text="@{contributor.rankOrder}"
+            android:textAppearance="@style/TextAppearance.DroidKaigi.Subtitle2"
```
takahirom: The reference may have been wrong, but you can write it like this: 👍 
```
                    android:textAppearance="?textAppearanceSubtitle2"
```

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionDetailSpeakerItem.kt
```diff
@@ -52,23 +51,13 @@ class SessionDetailSpeakerItem @AssistedInject constructor(
...
-                context.resources,
-                R.drawable.ic_person_outline_black_32dp,
-                null
-            )?.apply {
-                setTint(
-                    context.getThemeColor(R.attr.colorOnBackground)
-                )
-            }
-        }?.also {
+        val placeholder = ProfilePlaceholderCreator.create(context)?.also {
```
takahirom: 👏

feature/contributor/src/main/res/layout/fragment_contributors.xml
```diff
@@ -5,34 +5,31 @@
     xmlns:tools="http://schemas.android.com/tools"
     >
 
-    <androidx.constraintlayout.widget.ConstraintLayout
+    <FrameLayout
```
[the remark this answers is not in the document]
mkano9: It's mostly for simplicity. I don't foresee any changes/features on this screen that will need ConstraintLayout later. Also a little bit for performance in theory, but I know it won't matter.

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/ContributorsFragment.kt
```diff
@@ -18,25 +19,34 @@ import io.github.droidkaigi.confsched2020.App
...
     }
+    @Inject lateinit var systemViewModelProvider: Provider<SystemViewModel>
+    private val systemViewModel: SystemViewModel by assistedActivityViewModels {
+        systemViewModelProvider.get()
+    }
+    @Inject lateinit var contributorItemFactory: ContributorItem.Factory
 
+    private var binding: FragmentContributorsBinding by autoCleared()
     private var progressTimeLatch: ProgressTimeLatch by autoCleared()
+    private val groupAdapter = GroupAdapter<ViewHolder<*>>()
```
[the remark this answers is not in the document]
mkano9: Nice extension, I will use that

mkano9:
@takahirom Thanks for the second round of review. It's ready for another review :)

mkano9:
Thanks a bunch again for reviewing :)

PR #487, opened 2020-01-20
Apply review comments on #469 (Misuse groupie on SessionDetailFragment)

tfandkusu:
## Issue

none

## Overview (Required)
- Apply review comments on #469

## Degrade

When apply review comments, two issues appear. 

- When click favorite session button, entire screen show fade in effect.
- When click favorite session button, opened session description is closed.

To resolve first problem, I use ItemAnimator (supportsChangeAnimations = false). But I can not resolve it. Second problem needs time to understand implementation. So could someone fix in other PR or tell me how to fix it ?

PR #501, opened 2020-01-21
Disable fade in effect in SessionDetailFragment when click session favorite button

tfandkusu:
## Issue

No

## Overview (Required)
- Disable fade in effect in the SessionDetailFragment when click session favorite button.
- It is a degrade of #487

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionDetailDescriptionItem.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionDetailDescriptionItem.kt
@@ -15,6 +15,7 @@ import androidx.core.view.doOnPreDraw
 import androidx.transition.TransitionManager
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
+import com.xwray.groupie.Item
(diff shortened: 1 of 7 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionDetailDescriptionItem.kt
```diff
@@ -23,7 +23,7 @@ import io.github.droidkaigi.confsched2020.session.databinding.ItemSessionDetailD
 class SessionDetailDescriptionItem @AssistedInject constructor(
     @Assisted private val session: Session
 ) :
-    BindableItem<ItemSessionDetailDescriptionBinding>() {
+    BindableItem<ItemSessionDetailDescriptionBinding>(1L) {
```
takahirom: Thank you for checking behavior 👍 
My recommendation for this problem.

```
override fun isSameAs(other: Item<*>?): Boolean = other is SessionDetailDescriptionItem
```

tfandkusu: Thank you for telling Groupie usage.
I applied this comment.

PR #504, opened 2020-01-21
Added editorconfig to disable the ktlint rule for import ordering

### LINKED ISSUE #503: Import ordering rule of ktlint should be disabled due to the difference between AS and ktlint
**Overview (Required)**
- Sometimes Ktlint reports issues based on the rule but AndroidStudio's optimize imports feature cannot fix them, unfortunately.

Please read https://youtrack.jetbrains.com/issue/KT-10974 and https://github.com/pinterest/ktlint/issues/527 for more details.

**Steps To Reproduce**

You can see this problem in MainActivity

**Expected behavior**

Issues should be reported based on the same strategy but it's impossible. So we would like to see no issues with import orders.

**Screenshots**
N/A

**Env:**
N/A

**Additional context**
N/A

jmatsu:
## Issue
- close #503

## Overview (Required)
- Follow the expected behaviour that #503 noted. The reason why we need to do it is also described.
- To check the behaviour, I ran `./gradlew ktlintFormat` and no diffs are generated as expected.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/.ci/.editorconfig
+++ b/.ci/.editorconfig
@@ -1,17 +1,1 @@
-root = true
-
-[*]
-indent_style = space
-indent_size = 4
-charset = utf-8
-trim_trailing_whitespace = true
-insert_final_newline = true
-
-[*.Dangerfile]
-indent_size = 2
-
-[*.Gemfile]
-indent_size = 2
-
-[*.Rakefile]
-indent_size = 2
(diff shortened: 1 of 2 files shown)


takahirom:
👍

PR #509, opened 2020-01-21
Build failed on building for real device

ry-itto:
## Issue
I'm sorry, I didn't create some issue.
- close #ISSUE_NUMBER

## Overview (Required)
This is for iOS project.
- Build failed on building for real device
  - The reason for this problem is we use always `iosX64` module.
- To generate `xcconfig` files by using `xcconfig-extractor`.
  - The reason why I don't add `xcconfig-extractor` dependency to `Mintfile` and `README.md` is we don't need to generate `xcconfig` files many times.

## Links
- [xcconfig-extractor](https://github.com/toshi0383/xcconfig-extractor)
- [XcodeGen - Configuring build settings](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/Usage.md#configuring-build-settings)
- [Kotlin/Native Multiplatform プロジェクトで Android/iOS 向けの共通ライブラリを作る](https://qiita.com/irgaly/items/eebd1273b38e1e671fcf#ios-%E5%90%91%E3%81%91-framework-%E3%82%92%E3%83%93%E3%83%AB%E3%83%89%E3%81%99%E3%82%8B)


takahirom:
LGTM 👍

PR #511, opened 2020-01-21
Remove binding property from Fragment property

kubode:
## Overview (Required)

Although `autoCleared()` solves memory leaks of `binding` property, binding does not have to be a `Fragment`'s property in the first place.
Therefore, I'd like to propose a way to setup `binding` with the extension inside a class.

If you accept this proposal, I'm going to continue refactoring other Fragments.


feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SearchSessionsFragment.kt
```diff
@@ -100,10 +94,14 @@ class SearchSessionsFragment : DaggerFragment() {
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        FragmentSearchSessionsBinding.bind(view).setup()
+    }
+
+    private fun FragmentSearchSessionsBinding.setup() {
```
takahirom: I agree with the overall policy!
This is my preference, but since there is not much processing other than binding, I prefer not to split this method.

kubode: Thanks for agreeing.
Do you mean refactor like the following?
```kotlin
    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
        super.onViewCreated(view, savedInstanceState)
        val binding = FragmentSearchSessionsBinding.bind(view)
        ...
        binding.searchSessionRecycler.adapter = groupAdapter
        ...
    }
```

kubode:
Could you please take a look again?

takahirom:
Thanks! We will try it!🙏

PR #526, opened 2020-01-21
More options for dark theme

### LINKED ISSUE #466: Behave consistent around dark theme
## Kind (Required)
- Proposal / Discussion

 ## Overview (Required)
- Currently app is themed with system settings (related: #465) on launch. It seems to behave like `MODE_NIGHT_FOLLOW_SYSTEM` so switching system dark settings updates switch in app as well. 
- However, once check/uncheck the dark theme switch, it behaves as MODE_NIGHT_YES/MODE_NIGHT_NO. At that time, switching system dark settings no longer updates switch in app.
- It is good to add `MODE_NIGHT_FOLLOW_SYSTEM` so that user can choose fixed(dark/light) or system dependent. 

How about it?

 ## Links	
- https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94

### LINKED ISSUE #560: Dark theme chagne when tap Setting in Navigation drawer.
**Overview (Required)**
- The dark theme of the app is turned off when tap setting in menu, if device dark theme is set.

**Steps To Reproduce**
Please watch gif attached to screenshots.

**Expected behavior**
Dark theme change when tap dark theme switch.

**Screenshots**
device-2020-01-23-164504

**Env:**
 - Device: Pixel3xl
 - OS: 10
 - deploygate #167

ytRino:
## Issue
- close #466 close #560 

## Overview (Required)
- Follow system and battery option for dark theme.
    - Before once user switched dark theme in app, they no longer back to system dependent mode.
- Default value is "System default" (API 29 and above) or "When  battery saver is on" (API 28 and below).
    - App become dark when system dark setting is on (API 29 and above) or battery saver is on (all).
    - Selecting "Dark" or "Light" make app ignores both system dark setting and battery saver setting.

## Problem
- If we put preference's string resource in preference module, `App` cannot reference it. On the other hands, if we put them in base module, preference fragment needs to use them as another R class while R class of preference xml is in preference package. ([I did latter](https://github.com/DroidKaigi/conference-app-2020/pull/526/files#diff-f50804c80f18dda93732b8485c6fe46cR17))
- It is backed by preference screen's SharedPreferences. `App` references that SharedPreferences directly.

## Links
- https://medium.com/androiddevelopers/appcompat-v23-2-daynight-d10f90c83e94

## Screenshot

Before | After (API >= 29) | After (API < 29)
:--: | :--: | :--:
 |  |

### DIFF
--- a/android-base/src/main/java/io/github/droidkaigi/confsched2020/App.kt
+++ b/android-base/src/main/java/io/github/droidkaigi/confsched2020/App.kt
@@ -53,6 +53,7 @@ open class App : DaggerApplication(), AppComponentHolder {
                 .getString(DARK_THEME_KEY, getString(R.string.pref_theme_value_default))) {
                 getString(R.string.pref_theme_value_dark) -> AppCompatDelegate.MODE_NIGHT_YES
                 getString(R.string.pref_theme_value_light) -> AppCompatDelegate.MODE_NIGHT_NO
+                getString(R.string.pref_theme_value_battery) -> AppCompatDelegate.MODE_NIGHT_AUTO_BATTERY
                 else -> AppCompatDelegate.MODE_NIGHT_FOLLOW_SYSTEM
             }
         AppCompatDelegate.setDefaultNightMode(nightMode)
--- a/android-base/src/main/java/io/github/droidkaigi/confsched2020/App.kt
+++ b/android-base/src/main/java/io/github/droidkaigi/confsched2020/App.kt
@@ -50,7 +50,7 @@ open class App : DaggerApplication(), AppComponentHolder {
     private fun setupNightMode() {
         val nightMode =
             when (PreferenceManager.getDefaultSharedPreferences(applicationContext)
-                .getString(DARK_THEME_KEY, getString(R.string.pref_theme_value_default))) {
+                .getString(DARK_THEME_KEY, getString(R.string.pref_theme_default))) {
                 getString(R.string.pref_theme_value_dark) -> AppCompatDelegate.MODE_NIGHT_YES
                 getString(R.string.pref_theme_value_light) -> AppCompatDelegate.MODE_NIGHT_NO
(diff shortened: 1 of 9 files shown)


takahirom:
You can create a class like the Prefs class in the androidcomponents module. 👀

ytRino:
Currently PrefrencesViewModel does not do any useful work 😭 but I think it is enough for now.

takahirom:
Can you merge from master? 🙏

android-base/src/main/java/io/github/droidkaigi/confsched2020/App.kt
```diff
@@ -12,11 +11,12 @@ import io.github.droidkaigi.confsched2020.di.AppComponent
 import io.github.droidkaigi.confsched2020.di.AppComponentHolder
 import io.github.droidkaigi.confsched2020.di.createAppComponent
 import io.github.droidkaigi.confsched2020.image.CoilInitializer
+import io.github.droidkaigi.confsched2020.util.Prefs
 import timber.log.LogcatTree
 import timber.log.Timber
 
 open class App : DaggerApplication(), AppComponentHolder {
-    private val SWITCH_DARK_THEME_KEY = "switchDarkTheme"
+
```
takahirom:
```suggestion
```

model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/NightMode.kt
```diff
@@ -0,0 +1,8 @@
+package io.github.droidkaigi.confsched2020.model
+
+enum class NightMode {
```
takahirom: 🥇

ytRino:
oops I merged master then big diff have generated by line break mode change 😢

ytRino:
Don't know if it is right way but having this and solved.
```
git config --global core.autocrlf input
```

takahirom:
Thanks!

PR #527, opened 2020-01-21
Change iOS deployment target

woxtu:
## Issue
- None

## Overview (Required)
- Current iOS deployment target is too high, so I changed it to 12.0.
- I think it's enough to support the latest 2 versions.

## Links
- https://developer.apple.com/support/app-store/

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ios-base/project.yml
+++ b/ios-base/project.yml
@@ -5,7 +5,7 @@ targets:
     DroidKaigi 2020:
         type: application
         platform: iOS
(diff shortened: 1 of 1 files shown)


ry-itto:
Please fix `Podfile` too. 🙏

woxtu:
@ry-itto Thank you for pointing that out 🙏

PR #521, opened 2020-01-21
Introduce SwiftLint

yuta24:
## Issue

N/A

## Overview (Required)
- Enable SwiftLint for Swift code
- Run lint before build

## Screenshot
Before | After
:--: | :--:
 |

PR #529, opened 2020-01-21
Introduce SwiftFormat

woxtu:
## Issue
- None

## Overview (Required)
- Introduce SwiftFormat for good development
- Added minimum rules for now, so feel free to modify it
- Referred #521 to create this PR, thanks @yuta24!

## Links
- https://github.com/nicklockwood/SwiftFormat
- https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ios-base/scripts/format.sh
+++ b/ios-base/scripts/format.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+git diff --name-only | grep -e '\(.*\).swift$' | while read filename; do
+  # ref: https://github.com/yonaskolb/Mint/issues/112
+  mint run swiftformat swiftformat "$(git rev-parse --show-toplevel)/$filename"
+done
--- a/ios-base/scripts/format.sh
+++ b/ios-base/scripts/format.sh
@@ -2,5 +2,5 @@
 
 git diff --name-only | grep -e '\(.*\).swift$' | while read filename; do
   # ref: https://github.com/yonaskolb/Mint/issues/112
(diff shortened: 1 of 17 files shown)


ios-base/scripts/format.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+git diff --name-only | grep -e '\(.*\).swift$' | while read filename; do
+  # ref: https://github.com/yonaskolb/Mint/issues/112
+  mint run swiftformat swiftformat "$(git rev-parse --show-toplevel)/$filename"
```
ry-itto: nits.
use `$SRCROOT/../$filename` instead of `$(git rev-parse --show-toplevel)/$filename` is also ok 😄

woxtu: I'm not sure which is better 🤔

ry-itto: If we match writing style about this, I think you should use `$SRCROOT/../$filename`, because it is already used by `lint.sh`. 👍

yuta24: I think `$(git rev-parse --show-toplevel)/$filename` better than `$SRCROOT/../$filename`. If script using `$(git rev-parse --show-toplevel)/$filename`, it's executable via commandline.

However, since it is rarely executed from the command line, I think that either is good.

woxtu: Thank you both! I decided to use `$SRCROOT` to match others.

takahirom:
Thanks!

PR #540, opened 2020-01-22
Fix time label is gone if there are over 2 sessions had same start time

gen0083:
## Issue
- no issue but bug fix

## Overview (Required)
- if there are over 2 sessions had same start time, time label is gone

## Screenshot
Before | After
:--: | :--:
time_label_before 2020-01-22 22_01_27 | time_label_after 2020-01-22 21_53_54


takahirom:
LGTM!👍

PR #461, opened 2020-01-20
#104 Issue resolved

yukiyuki961:
## Issue
- close #104 

## Overview (Required)
- The colors of the icons on the bottom sheet and the navigation bar are similar, making it difficult to see.

## Notes
- I really wanted to get a swipe verdict for `ButtomAppBar` . Upon examination, I found no Listener that could be used in `BottomAppBar` .
- Looking for other alternatives, `AppbarLayout` seemed to have an `AppBarLayout.OnOffsetChangedListener`so I could do it.
But `AppbarLayout` didn't seem to be able to insert a `FloatingActionButton` .
Therefore, I decided not to make a swipe decision and set translucent white.

## Screenshot
Before | After
:--: | :--:
 | 
image | image

PR #542, opened 2020-01-22
#104 Fixed navigationbar on detail screen.

### LINKED ISSUE #104: When BottomSheet is expanded, top left corner is not rounded
## Kind (Required)

Choose the suitable template if exists.

- Other

 ## Overview (Required)

- Title says it all.  

 ## Screenshot

valjapan:
Since I did not push my master branch, I will push again. 
Here is the previous pull request.(#461)

## Issue
- close #104 

## Overview (Required)
- The colors of the icons on the bottom sheet and the navigation bar are similar, making it difficult to see.

## Notes
- @jmatsu says, I used `systemUiVisibility`. Thank you for the advice.
- As a result, the navigation is not transparent only within this fragment.

## Screenshot
Before | After | After(now)
:--: | :--: | :--:
 | 
image | image | image


feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
```diff
@@ -99,6 +100,10 @@ class SessionDetailFragment : DaggerFragment() {
             container,
             false
         )
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            activity?.window?.decorView?.systemUiVisibility =
```
takahirom: We are using SystemUiManager class. can you check it? 🙏

jmatsu (changes requested):
Thank you so much! The approach looks good. Could you please address only one point which was from takahirom? ref: https://github.com/DroidKaigi/conference-app-2020/pull/542/files#r370191036

yukiyuki961:
Sorry for the late resolution of the issue.
I'm struggling to call SystemUIManagar on android-base.
Let me do my best a little more.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionDetailFragment.kt
yukiyuki961: @takahirom 
sorry. I was busy working and couldn't help.
Unfortunately I couldn't call SystemUiManager.
Could you give me any advice?

takahirom: Of course! 
How about moving SystemUiManager to android corecomponet?

yukiyuki961: @takahirom 
This is solved using SystemUiManager! Thank you for the advice!
 commit a08be2aaf8c3692050dcfd3c88c0ff2ecdbe4c3d

PR #545, opened 2020-01-22
bottom sheet behavior request force update

### LINKED ISSUE #520: The front layer of the backdrop (session list) becomes empty and stays in the middle of the screen
**Overview (Required)**
- When opening search screen from session list screen
  and come back to session list,
- The front layer of the backdrop (session list) becomes empty
  and it stays in the middle of the screen

I have a screen recording so that you can see what I get in detail.

**Steps To Reproduce**

1. Go to the search screen
2. Click back arrow in the search screen
3. See the backdrop stays in the middle
4. And the session list is invisible

**Expected behavior**
- The backdrop position should be at the bottom (where it was right before opening search screen)
- Session list should be visible

**Screenshots**
ezgif-6-57037a1fdbee

**Env:**
 - Device: Pixel 4 XL
 - OS: Android 10
 - Version: DeployGate #139 / 1.0.5 (1000500)

**Additional context**
This is possibly related to https://github.com/DroidKaigi/conference-app-2020/issues/505 as I see the crash and have a crash report having the same stack trace.

hirotakaakita:
## Issue
- close #520

## Overview (Required)
- Even if peekHeight was changed, the height of BottomSheetBehavior was not recalculated, so it was recalculated.
- requestLayout() uses onLayoutChild() because the phenomenon was not resolved.
- To save computational resources, STATE_COLLAPSED has been added to work only for issues

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
@@ -48,7 +48,7 @@ class SessionsFragment : DaggerFragment() {
     private var binding: FragmentSessionsBinding by autoCleared()
     private lateinit var overrideBackPressedCallback: OnBackPressedCallback
 
-    private val sessionSheetBehavior: BottomSheetBehavior<*>
+    private val sessionSheetBehavior: BottomSheetBehavior<View>
         get() {
             val layoutParams = binding.sessionsSheet.layoutParams as CoordinatorLayout.LayoutParams
             val behavior = layoutParams.behavior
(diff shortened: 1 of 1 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/SessionsFragment.kt
```diff
@@ -113,6 +113,12 @@ class SessionsFragment : DaggerFragment() {
         binding.sessionsSheet.doOnApplyWindowInsets { _, insets, _ ->
             sessionSheetBehavior.peekHeight =
                 insets.systemWindowInsetBottom + initialPeekHeight + gestureNavigationBottomSpace
+            if (sessionSheetBehavior.state == BottomSheetBehavior.STATE_COLLAPSED)
```
takahirom: Thanks! 👍 
Can you add a comment? about this workaround 🙏

hirotakaakita: Comment added 👍

takahirom:
LGTM🙏

PR #565, opened 2020-01-23
fix announcement link text

### LINKED ISSUE #525: <a>tag text is not link
**Overview (Required)**
- <a>tag text is not link

**Expected behavior**
I this <a>tag text is link text

**Screenshots**
link_lost

**Env:**
 - Device: SHV43
 - OS: 9.0
 - App Version or the branch 

**Additional context**
Any other notes

numa08:
## Issue

- close #525 

## Overview (Required)

- The TextView in announcement recyclerview analyzes link text, and it doesn't interpret html tag

## Links
-

## Screenshot
Before | After
:--: | :--:
before | after

### DIFF
--- a/feature/announcement/src/main/res/layout/item_announcement.xml
+++ b/feature/announcement/src/main/res/layout/item_announcement.xml
@@ -49,7 +49,7 @@
             android:layout_marginTop="14dp"
             android:textColor="?colorOnBackground"
             android:textColorLink="@color/design_default_color_secondary"
-            android:autoLink="all"
+            android:autoLink="web"
             app:layout_constraintEnd_toEndOf="@+id/guideline_end"
             app:layout_constraintStart_toStartOf="@+id/announcement_title"
             app:layout_constraintTop_toBottomOf="@+id/announcement_title"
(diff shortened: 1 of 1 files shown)


feature/announcement/src/main/res/layout/item_announcement.xml
```diff
@@ -48,6 +48,8 @@
             android:layout_height="wrap_content"
             android:layout_marginTop="14dp"
             android:textColor="?colorOnBackground"
+            android:textColorLink="@color/design_default_color_secondary"
+            android:autoLink="all"
```
takahirom: Can you use autolink=web? because probably we don't use a phone number, mail address 👍

jmatsu: I agreed. Phone numbers, mail addresses, and map addresses are needless for now.

By the way, *Linkify.MAP_ADDRESSES* (and also Linkify.ALL of course) may cause a crash on devices that do not have Webkit. Android devices require `WebView#findAddress` to grep Geo URLs and throw a ClassNotFoundException if it's missing.

numa08: I resolved it !

takahirom:
Thanks!!!

PR #569, opened 2020-01-24
Current session time scroll

### LINKED ISSUE #150: Show current time session when loading session first in Fragment

takahirom:
## Issue
- close #150

## Overview (Required)
- Scroll to current session
- Refactor around session list

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/SessionList.kt
```diff
@@ -0,0 +1,52 @@
+package io.github.droidkaigi.confsched2020.model
+
+data class SessionList(private val sessions: List<Session>) : List<Session> by sessions {
```
takahirom: I will create issue for creating test 🙇

takahirom:
scroll

Issue #494, opened 2020-01-21
iOS App sponsor view


ry-itto:
## Kind (Required)
- Proposal / Discussion

 ## Overview (Required)
- A view that shows sponsors.

 ## Links	
- [Design](https://www.figma.com/file/4r9becvhDy3GfXaXex8E8d/App?node-id=22%3A507)

yaslab:
🙋‍♂️ I will try this! I'm interested in Material Components for iOS.

satsukies:
@yaslab Thanks! We assign this issue to you!

PR #573, opened 2020-01-24
[iOS] Improve asynchronous loading of UIImageView

yaslab:
## Issue
- N/A

## Overview (Required)
- I'm implementing issue #494. This issue's screen has image view for each collection view cells.
- I'm trying to use [this method](https://github.com/DroidKaigi/conference-app-2020/blob/884cf4e772f176fa449aac797b0a8c054852cb74/ios-base/DroidKaigi%202020/Extentions/UIImageView%2BEx.swift#L4) for image loading, but this method calls an HTTP loading every time the collection view reuses the cell.
- I added the following two features.
    - (1) Canceling image loding task: To cancel previous image loading. e.g. when reusing cells.
    - (2) Caching image: To avoid loading already loaded images again.

## Links
- [SwiftでObjective-Cの黒魔術ってどうなった？](https://qiita.com/fmtonakai/items/e9036dec4af2609b5715)
- [NSBlockOperationで手軽にキャンセル処理](https://qiita.com/ninjinkun/items/0fe8548f1debd783a04a)

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ios-base/DroidKaigi 2020/Extentions/UIImageView+Ex.swift
+++ b/ios-base/DroidKaigi 2020/Extentions/UIImageView+Ex.swift
@@ -1,20 +1,73 @@
 import UIKit
 
 extension UIImageView {
-    func loadImage(url: URL?) {
+    class Cache {
+        private var cache = [String: UIImage]()
+
+        fileprivate func setImage(_ image: UIImage, for key: URL) {
+            cache[key.absoluteString] = image
+        }
+
+        fileprivate func image(for key: URL) -> UIImage? {
+            return cache[key.absoluteString]
+        }
+    }
+
+    private static let imageLoadingQueue = OperationQueue()
+
+    private static var imageLoadingOperationKey = 0
+
+    private var imageLoadingOperation: Operation? {
+        get {
+            return objc_getAssociatedObject(
+                self,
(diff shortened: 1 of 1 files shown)


ios-base/DroidKaigi 2020/Extentions/UIImageView+Ex.swift
```diff
@@ -1,20 +1,73 @@
 import UIKit
 
 extension UIImageView {
-    func loadImage(url: URL?) {
+    class Cache {
+        private var cache = [String: UIImage]()
```
yanamura: I think it's better to  add cache limit or  use NSCache.

yuta24:
Great 👍
But, What about using existing libraries such as [Kingfisher](https://github.com/onevcat/Kingfisher) or [Nuke](https://github.com/kean/Nuke) ?

ios-base/DroidKaigi 2020/Extentions/UIImageView+Ex.swift
yaslab: Good catch! I completely forgot about `NSCache`...

yaslab:
Exactly so! I will consider using libraries.

Issue #552, opened 2020-01-23
Add date for first session of day when SessionPage.Favorite


takahirom:
## Kind (Required)

- Improvement

 ## Overview (Required)

- 
![image](https://user-images.githubusercontent.com/1386930/72946924-c28a9300-3dc3-11ea-8866-460f279f01e6.png)

 ## Links	

-

PAGMF:
🙋‍♂

CaptainPAG:
🙋‍♂

roana0229:
Assigned 👍

FujiKinaga:
It seems to be fixed by #574 . So I'll close this PR👋
But, We have a problem in #752.
We are looking forward to someone's contribution!

PR #574, opened 2020-01-24
Favorite session date text

CaptainPAG:
## Issue
- #552

## Overview (Required)
- Add date text only favorite session list

## Screenshot
Before | After
:--: | :--:
 |


feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -15,14 +15,15 @@ import io.github.droidkaigi.confsched2020.util.AndroidRAttr
 
 class SessionsItemDecoration(
     private val adapter: GroupAdapter<*>,
-    private val context: Context
+    private val context: Context,
+    private val isShowDate: Boolean
```
jmatsu: `isShowDate` looks ambiguous because it's shadowed due to [here](https://github.com/DroidKaigi/conference-app-2020/pull/574/files#diff-07d7092decb1657324b9691c9a54e57cR89) and `isShowDateText` is also defined.

How about `visibleSessionDate` for here?

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -44,7 +48,14 @@ class SessionsItemDecoration(
             if (position == RecyclerView.NO_POSITION) return
 
             val sessionItem = adapter.getItem(position) as SessionItem
-            val startTimeText = calcTimeText(position, view)
+            val isShowDateText = if (position > 0) {
```
jmatsu: The following looks clearer.

```suggestion
            val shouldShowDateText = visibleSessionDate  && if (position > 0) {
```

model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/Session.kt
```diff
@@ -14,7 +14,7 @@ sealed class Session(
     open val room: Room,
     open val isFavorited: Boolean
 ) {
-    val startDayText by lazy { startTime.toOffset(defaultTimeZoneOffset()).format("yyyy.M.d") }
+    val startDayText by lazy { startTime.toOffset(defaultTimeZoneOffset()).format("M/d") }
```
jmatsu: Make sense 😄

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -54,34 +65,93 @@ class SessionsItemDecoration(
...
+
+        return StartDateTimeText(dateText, startTimeText)
     }
 
+    private data class StartDateTimeText(
+        val dateText: DateText?,
+        val startTimeText: StartTimeText
+    )
+
+    private data class DateText(
```
jmatsu: Is this almost the same class as `StartTimeText`? then it would be enough to have only one class like `PositionalText`.

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -54,34 +65,93 @@ class SessionsItemDecoration(
...
+                startDateTimeText.startTimeText.value,
+                startDateTimeText.startTimeText.positionX,
+                startDateTimeText.startTimeText.positionY,
                 textPaint
             )
         }
     }
 
-    private fun calcTimeText(position: Int, view: View): StartTimeText {
+    private fun calcDateTimeText(
```
jmatsu: Great work. 👀

jmatsu: I felt it would be nicer if this method could be split into `calcTimeText` and `calcDateText` with offset calculation when I glanced, but it would be kinda heavy work. So it's okay but we are welcoming additional improvements anytime.

takahirom:
@CaptainPAG If you have any questions. Please tell me 🙏

FujiKinaga:
@CaptainPAG 
I work on this PR to release this feature!😄
Thank you for your contribution!

feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -54,35 +56,102 @@ class SessionsItemDecoration(
...
+        if (position <= 0) {
+            return true
+        }
+        val currentSessionItem = adapter.getItem(position) as SessionItem
+        val lastSessionItem = adapter.getItem(position - 1) as SessionItem
+        val isSameDateWithLastItem =
+            currentSessionItem.startSessionDate() == lastSessionItem.startSessionDate()
+        return when {
+            isSameDateWithLastItem -> false
+            else -> true
```
FujiKinaga: When multiple same time's favorite session has, date time text will dismiss on scrolling one item height in this code.😭
But, I was not be able to fix this problem. I'll issue for the problem once.
fps_30

model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/SessionPage.kt
```diff
@@ -26,6 +26,8 @@ sealed class SessionPage : AndroidParcel {
 
     abstract val title: String
 
+    fun visibleSessionDate() = this is Event || this is Favorite
```
takahirom: 👍

takahirom:
LGTM 👍

PR #581, opened 2020-01-24
Use property 'shouldScroll'

### LINKED ISSUE #424: Sometimes, when returning from session details, the scroll position of the session list is reset
**Overview (Required)**
- Sometimes, when returning from session details, the scroll position of the session list is reset

**Steps To Reproduce**

1. Go to session details.
2. Return session list.
3. Sometimes,reload and reset position of session list.

**Expected behavior**
Don't reload and reset position.

**Screenshots**
N/A

**Env:**
 - Device: Xperia 10
 - OS: 9.0
 - master

numa08:
## Issue
- close #424 

## Overview (Required)

`SessionsList.currentSessionIndex` had returned `0` in spite of the session dose not started.

## Screenshot
Before | After
:--: | :--:
 
before|after

### DIFF
--- a/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/SessionList.kt
+++ b/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/SessionList.kt
@@ -20,12 +20,18 @@ data class SessionList(private val sessions: List<Session>) : List<Session> by s
     }
 
     val currentSessionIndex by lazy {
-        val lastFinished = indexOfLast { it.isFinished }
-        if (size - 1 == lastFinished) {
-            // if it is last we don't use it
-            -1
-        } else {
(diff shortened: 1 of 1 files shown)


takahirom:
👀

takahirom:
Thank you for finding bug cause!!!
Probably we can fix in SessionList like this 👍 

```kotlin
    val currentSessionIndex by lazy {
        val lastFinished = indexOfLast { it.isFinished }
        if (lastFinished == -1) {
            // if session does not started we don't use it
            -1
        } else if (size - 1 == lastFinished) {
            // if it is last we don't use it
            -1
        } else {
            lastFinished + 1
        }
    }
```

numa08:
Thanks ! I'll try it !!

takahirom:
Thanks! LGTM!

PR #582, opened 2020-01-24
Save state which description show full text

tfandkusu:
## Issue

No

## Overview (Required)
- On SessionDetailFragment, when description shows full text and click favorite session button, description back to ellipsis. 
- When user rotates device, description back to ellipsis, too.
- So this PR saves this state.

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Session detail structure is changed.
And that issue is assigned and fixed 😭 
https://github.com/DroidKaigi/conference-app-2020/issues/325
I look forward to your next contribution!

PR #585, opened 2020-01-25
[iOS] Replace value of surfaceColor

yaslab:
## Issue
- N/A

## Overview (Required)
- I replaced value of `surfaceColor` with `.white`. (Refer to [Android implementation](https://github.com/DroidKaigi/conference-app-2020/blob/master/corecomponent/androidcomponent/src/main/res/values/themes.xml#L25))
- And set `onSurfaceColor` to `.black`.
- (The reason for this change is that I want to use `surfaceColor` as `.white` in [Sponsor screen](https://github.com/DroidKaigi/conference-app-2020/issues/494))

## Links
- N/A

## Screenshot
Before | After
:--: | :--:
 |


ry-itto:
Thanks! LGTM 😄

PR #602, opened 2020-01-25
Expand ripple effect for speaker

phicdy:
## Issue

Nothing

## Overview (Required)
- We can tap all of the line of speaker in session detail, but ripple effect is only for speaker icon and name 

## Links

Nothing

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
LGTM 👍

PR #616, opened 2020-01-26
Add retry on Contributors screen

mkano9:
## Issue
- No Issue filed
- Follow-up from #115 

## Overview (Required)
- Allow the user to retry reloading the data when the API call fails.

Before:
- Problem 1: When no cache exists and the API call fails, there's no way to reload the Contributors screen unless you exit and re-enter the Contributors screen.
- Problem 2: The loading indicator gets dismissed as soon as the repository loads data from DB regardless of whether cache exists or not. The app then starts an API call which could make the user wait for some time on a blank screen without any loading indicator until the API call completes.

After:
- When the API call fails, if there's no cache to show any data, then show a retry button. If cache exists, then show the retry option on the snackbar while displaying the current data.
- Loading indicator continues to be displayed until either real data is retrieved or an error occurs. 

## Screenshot
Note:
- Airplane mode is enabled in the beginning of the videos to simulate network errors.
- Added a delay of 3s to the API calls to simulate slow network connection.

Before | After
:--: | :--:
 | 

Error when contributors cache exists:

### DIFF
--- a/feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
+++ b/feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
@@ -1,16 +1,21 @@
 package io.github.droidkaigi.confsched2020.contributor.ui.viewmodel
 
 import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.asLiveData
 import androidx.lifecycle.liveData
+import androidx.lifecycle.viewModelScope
 import io.github.droidkaigi.confsched2020.ext.combine
+import io.github.droidkaigi.confsched2020.ext.dropWhileIndexed
 import io.github.droidkaigi.confsched2020.ext.toAppError
 import io.github.droidkaigi.confsched2020.ext.toLoadingState
 import io.github.droidkaigi.confsched2020.model.AppError
 import io.github.droidkaigi.confsched2020.model.Contributor
 import io.github.droidkaigi.confsched2020.model.LoadState
+import io.github.droidkaigi.confsched2020.model.LoadingState
 import io.github.droidkaigi.confsched2020.model.repository.ContributorRepository
+import kotlinx.coroutines.launch
 import java.lang.Exception
 import javax.inject.Inject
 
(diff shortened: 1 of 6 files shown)


takahirom:
👀

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
```diff
@@ -33,28 +38,50 @@ class ContributorsViewModel @Inject constructor(
     private val contributorsLoadStateLiveData: LiveData<LoadState<List<Contributor>>> = liveData {
         emitSource(
             contributorRepository.contributorContents()
+                .dropWhileIndexed { index, value ->
```
takahirom: 👍

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
```diff
@@ -33,28 +38,50 @@ class ContributorsViewModel @Inject constructor(
...
+                }
                 .toLoadingState()
                 .asLiveData()
         )
 
         try {
             contributorRepository.refresh()
         } catch (exception: Exception) {
-            // We can show contributors with cache
+            reloadContributorsLiveData.value = LoadingState.Error(exception)
```
takahirom: It's a little difficult here, but I can't think of any better solution. 👀

feature/contributor/src/main/res/values-ja/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="retry_label">リトライ</string>
```
takahirom: Thanks 😄

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
mkano9: Yeah, I didn't like this, but I didn't have better ideas. Introducing another LiveData just for this error scenario would've been worse.

feature/contributor/src/main/java/io/github/droidkaigi/confsched2020/contributor/ui/viewmodel/ContributorsViewModel.kt
mkano9: TBH, I wish we had a mechanism set up where if cache doesn't exist, the repository would not emit :)

mkano9:
@takahirom Thank you again for the review!

PR #617, opened 2020-01-26
Add ripple effect to Read more button in the SessionDetailFragment

tfandkusu:
## Issue

none

## Overview (Required)
- Add ripple effect to `Read more` button in the omitted description text of SessionDetailFragment.
- Clickable area of `Read more` button becomes large for accessibility.

## Links

Current implementation uses ClickableSpan, but it has no solution to add ripple effect.

- https://stackoverflow.com/questions/44105423/add-click-ripple-to-clickable-clickablespan

So, New TextView of `Read more` is added and its coordinates are calculated by omitted description text's character coordinates.

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Sorry I will look this 🙇

tfandkusu:
I withdraw this PR because current implementation which clickable range is all of description text has no accessibility problem.

PR #619, opened 2020-01-26
Fix layout structure for lint issue of UseCompoundDrawables

### LINKED ISSUE #611: This tag and its children can be replaced by one <TextView/> and a compound drawable
## Kind (Required)

- Improvement

 ## Overview (Required)

- The lint say it.
- Can we do it?? 👀 
https://github.com/DroidKaigi/conference-app-2020/blob/ba799e8c11633a0d9e6d9b400961d0582cb0358e/feature/session/src/main/res/layout/item_session_detail_speaker.xml#L19

 ## Links	

-

hoyo:
## Issue
- close #611

## Overview (Required)
- We cannot replace to compound drawable, because the ImageView is used for navigation transition.
- I propose using ConstraintLayout for avoid the lint warning message.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/res/layout/item_session_detail_speaker.xml
+++ b/feature/session/src/main/res/layout/item_session_detail_speaker.xml
@@ -5,8 +5,7 @@
     xmlns:tools="http://schemas.android.com/tools"
     >
 
-
-    <FrameLayout
+    <androidx.constraintlayout.widget.ConstraintLayout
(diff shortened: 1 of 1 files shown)


takahirom:
> ImageView is used for navigation

🙆 
Thanks! LGTM!!

PR #623, opened 2020-01-26
Fix search auto entered

### LINKED ISSUE #596: Fix search auto entered
## Kind (Required)

- Improvement

 ## Overview (Required)

-  We can not use completion in Android 10
search_entered

 ## Links	

-

sekiya9311:
## Issue
- close #596 

## Overview (Required)
- The cause of this bug is call `invalidateOptionsMenu()` in `toolbar.doOnApplyWindowInsets {}` when query text changed.
- So, changed place of call `invalidateOptionsMenu()` from `toolbar.doOnApplyWindowInsets {}`  to `toolbar.doOnLayout {}`
- We can use `toolbar.doOnNextLayout {}` insted of `toolbar.doOnLayout {}` and resolved too. But, I don't know which is better. 😥

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Thank you! You found the cause of this!

PR #638, opened 2020-01-28
Make session's text selectable

wada811:
## Issue
- N/A

## Overview (Required)
- I want to translate session's text from English to Japanese.
  - Make following text selectable in session detail
    - Title
    - Description
    - Target

## Links
- N/A

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Thanks! That is useful!

PR #639, opened 2020-01-28
[room database] use withTransaction androidx extension method

satoshun:
## Overview (Required)
- withTransaction method handles Coroutine and Transaction appropriately. I think, withTransaction extension method better than current approach. 

## Links
- https://developer.android.com/reference/kotlin/androidx/room/package-summary#withtransaction

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Looks good!!! Thanks!!

PR #640, opened 2020-01-28
Fix session's text selectable

wada811:
## Issue
- N/A

## Overview (Required)
- "Make session's text selectable #638" is wrong 🙇 
  - I thought I has applied `textIsSelectable` to Session's title, but applied to time.
  - This PR fix it.

## Links
- #638

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
🙆

PR #648, opened 2020-01-29
Use ContentLoadingProgressBar instead of ProgressBar

### LINKED ISSUE #642: Use ContentLoadingProgressBar instead of ProgressBar with ProgressTimeLatch
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)

- ContentLoadingProgressBar is useful view.
- ContentLoadingProgressBar might be able to do the same as ProgressBar with ProgressTimeLatch
- Or, we dare to use ProgressBar because of some reason?

 ## Links	

- [ContentLoadingProgressBar](https://developer.android.google.cn/reference/kotlin/androidx/core/widget/ContentLoadingProgressBar)
- [ProgressTimeLatchとは何なのか](https://qiita.com/tsumuchan/items/38346528da5ebb8d8e64)

rmakiyama:
## Issue
- close #642

## Overview (Required)
- Use ContentLoadingProgressBar instead of ProgressBar
  - I left the ProgressTimeLatch for other view
- I think, progress shows in no time without setting gone visibility in xml as default. 
  - (not using min show time because not call show() method.)
  - Is this right behavior?（or case by case?）
  - ex) show progress in about screen without delay 500ms.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
👀

feature/staff/src/main/java/io/github/droidkaigi/confsched2020/staff/ui/StaffsFragment.kt
```diff
@@ -78,16 +76,12 @@ class StaffsFragment : Fragment(R.layout.fragment_staffs) {
...
-        }.apply {
-            loading = true
-        }
+        binding.progressBar.show()
 
         // This is the transition for the stagger effect.
         val stagger = Stagger()
         staffsViewModel.uiModel.observe(viewLifecycleOwner) { uiModel ->
-            progressTimeLatch.loading = uiModel.isLoading
+            with(binding.progressBar) { if (uiModel.isLoading) show() else hide() }
```
takahirom: How about creating an extension function? 🔄

rmakiyama: I think so nice! I create it! 🙌

rmakiyama:
I think this method name is not really good.... do you have any idea? 🙇

takahirom:
How about isShow?
var ContentLoadingProgressBar.isShow

takahirom:
LGTM 👍 🔄

PR #671, opened 2020-01-31
fit sponsor item design on dark theme

FujiKinaga:
## Issue
- N/A

## Overview (Required)
- When open sponsor screen on dark theme applied, card background visible by transparent.
    - Should always be set `colorOnPrimary`(white)
- Touch feedback color should use common color on both light and dark theme because card background color always white!
  - It seems change ripple and background color on these below in `CardView`.
    - `colorSurface` : background color
    - `colorPrimary` : ripple color on StateChecked is true by 20% alpha

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


feature/sponsor/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,13 @@
...
+<resources>
+
+    <style name="Widget.DroidKaigi.Sponsor.Base.Card" parent="Widget.MaterialComponents.CardView">
+        <item name="cardBackgroundColor">?colorOnPrimary</item>
+        <item name="cardCornerRadius">8dp</item>
+        <item name="cardElevation">3dp</item>
+    </style>
+
+    <style name="Widget.DroidKaigi.Sponsor.Card" parent="Widget.DroidKaigi.Sponsor.Base.Card">
+        <item name="rippleColor">@color/black_alpha_12</item>
```
FujiKinaga: I didn't know what color should be set 😅

takahirom:
LGTM 👍

PR #682, opened 2020-02-01
[suggest]Remove isSameAs instead of Item.id

koji-1009:
## Issue
- close N/A

## Overview (Required)
`com.xwray.groupie.Item.isSameAs` is this code.

```
    /**
     * Whether two item objects represent the same underlying data when compared using DiffUtil,
     * even if there has been a change in that data.
     * <p>
     * The default implementation compares both view type and id.
     */
    public boolean isSameAs(Item other) {
        if (getLayout() != other.getLayout()) {
            return false;
        }
        return getId() == other.getId();
    }
```

`getLayout() != other.getLayout()` result is same as `other is HogeItem` result.
So I delete that methods.

And, `getId()` is following.

```
    private static AtomicLong ID_COUNTER = new AtomicLong(0);
    private final long id;

    public Item() {
        this(ID_COUNTER.decrementAndGet());
    }

    protected Item(long id) {
        this.id = id;
    }

    /**
     * If you don't specify an id, this id is an auto-generated unique negative integer for each Item (the less
     * likely to conflict with your model IDs.)
     * <p>
     * You may prefer to override it with the ID of a model object, for example the primary key of
     * an object from a database that it represents.
     *
     * @return A unique id
     */
    public long getId() {
        return id;
    }
```

Therefore, I add unique id.
(If we don't use a same layout, we don't request the uniqueness.)

## Links
- [Item.java](https://github.com/lisawray/groupie/blob/master/library/src/main/java/com/xwray/groupie/Item.java)

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionDetailDescriptionItem.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2020/session/ui/item/SessionDetailDescriptionItem.kt
@@ -16,7 +16,6 @@ import androidx.core.view.doOnPreDraw
 import androidx.transition.TransitionManager
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
-import com.xwray.groupie.Item
 import com.xwray.groupie.databinding.BindableItem
 import io.github.droidkaigi.confsched2020.ext.getThemeColor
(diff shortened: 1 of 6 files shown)


koji-1009:
see #684

takahirom:
Can you merge from master for fxing CI?

koji-1009:
Thanks!
Now merge it!

koji-1009:
Sorry, I was overlooked `DividerItem` class.

PR #690, opened 2020-02-01
[iOS] Add App icons.

ry-itto:
## Issue
- N/A

## Overview (Required)
- Add App icons to iOS App.

## Screenshot
Before | After
:--: | :--:
 |


ry-itto:
This is only adding App icons to iOS App, so I will merge this with no reviews.

PR #699, opened 2020-02-02
Navigation to map when tapped access

### LINKED ISSUE #680: Add `Access` implementation in about screen
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)

- I'm thinking about launching Google Map with an implicit intent.
-  https://github.com/DroidKaigi/conference-app-2020/blob/7f29bcb303e6056a933b9d687b21c9804640f647/feature/about/src/main/java/io/github/droidkaigi/confsched2020/about/ui/AboutFragment.kt#L89

 ## Links	

-

youmitsu:
## Issue
- close #680 

## Overview (Required)
- On about page, navigation to map when tapped access button.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/about/src/main/java/io/github/droidkaigi/confsched2020/about/ui/AboutFragment.kt
+++ b/feature/about/src/main/java/io/github/droidkaigi/confsched2020/about/ui/AboutFragment.kt
@@ -1,5 +1,7 @@
 package io.github.droidkaigi.confsched2020.about.ui
 
+import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
 import android.view.View
 import androidx.core.view.updatePadding
@@ -27,6 +29,7 @@ import io.github.droidkaigi.confsched2020.ext.assistedViewModels
 import io.github.droidkaigi.confsched2020.system.ui.viewmodel.SystemViewModel
 import javax.inject.Inject
 import javax.inject.Provider
+
 
 class AboutFragment : Fragment(R.layout.fragment_about), Injectable {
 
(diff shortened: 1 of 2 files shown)


feature/about/src/main/java/io/github/droidkaigi/confsched2020/about/ui/AboutFragment.kt
```diff
@@ -87,7 +89,13 @@ class AboutFragment : Fragment(R.layout.fragment_about), Injectable {
                 aboutItemFactory.create(
                     getString(R.string.about_item_access)
                 ) {
-                    // TODO go access-page
+                    val intent = Intent(
+                        Intent.ACTION_VIEW,
+                        Uri.parse("geo:35.6219252,139.7190626?q=TOCビル")
+                    ).setPackage("com.google.android.apps.maps")
```
takahirom: I want to use implicit intent. So I think this is not needed 👀 
If you have any opinion please tell me.

feature/about/src/main/java/io/github/droidkaigi/confsched2020/about/ui/AboutFragment.kt
```diff
@@ -87,7 +89,13 @@ class AboutFragment : Fragment(R.layout.fragment_about), Injectable {
                 aboutItemFactory.create(
                     getString(R.string.about_item_access)
                 ) {
-                    // TODO go access-page
+                    val intent = Intent(
```
takahirom: Currently, we use `SystemViewModel` for launching intent. 🙏

youmitsu: Fixed.

feature/about/src/main/java/io/github/droidkaigi/confsched2020/about/ui/AboutFragment.kt
youmitsu: Fixed.

youmitsu:
@takahirom 
I fixed!
Could you check that again, please?

PR #703, opened 2020-02-02
Adjust line spacing of the description in about page

nashcft:
## Issue
- N/A

## Overview (Required)
- Set `lineSpacingExtra` to `8sp` like other multi-line text components for readability.

## Screenshot

Before | After
:--: | :--:
 |


mhidaka:
Thanks for your work! good catch!

PR #696, opened 2020-02-02
[iOS] Add blank event tab

clockvoid:
## Issue
- #657 

## Overview (Required)
- Add a blank event tab for the iOS app.
- I created a new SessionViewControllerType .event and adjust the raw number carefully.
- I created a new EventViewController because it's not Session and it doesn't have a filtering view.

## Screenshot
Before | After
:--: | :--:
 |

PR #708, opened 2020-02-03
Fix ios event tab

### LINKED ISSUE #657: iOS Add "Event" to tab
## Kind (Required)
- Improvement

 ## Overview (Required)
- add "Event" tab to FilterViewController.

 ## Links	
- https://www.figma.com/file/4r9becvhDy3GfXaXex8E8d/App?node-id=1279%3A1339

clockvoid:
## Issue
- close #657 

## Overview (Required)
- Delete EventViewController
  - In PR #696, I created EventViewController because I thought the events are not sessions. However, while reading the Android side source code, I noticed the events are obtained by the same endpoint as all sessions. So, I delete it and Implement the event tab by using if statement.
  - Please advise me which is the best solution.
- Implement filters for the event tab and the myplan tab.

## Links
- #696 

## Screenshot
Temporary, I use pictures filtered by backdrop instead of exhibition. Because, for now, there are sessions that are typed exhibition.

Before | After
:--: | :--:
 | 
 |


ry-itto:
@clockvoid 
> Temporary, I use pictures filtered by backdrop instead of exhibition. Because, for now, there are sessions that are typed exhibition.

This means session which has exhibition type is none?

ios-base/DroidKaigi 2020/Session/Views/SessionViewController.swift
```diff
@@ -66,7 +69,17 @@ final class SessionViewController: UIViewController {
...
             .map { sessions -> [Session] in
-                sessions.filter { Int($0.dayNumber) == self.type.rawValue }
+                switch self.type {
+                case .day1, .day2:
+                    return sessions.filter {
+                        Int($0.dayNumber) == self.type.rawValue
+                            && $0.room.roomType != Room.RoomType.exhibition
+                    }
+                case .event:
+                    return sessions.filter { $0.room.roomType == Room.RoomType.exhibition }
```
ry-itto: This one can write more simple 😄 
```swift
return sessions.filter { $0.room.roomType == .exhibition }
```

clockvoid:
@ry-itto 
Thank you for your reviewing.
Yes. For now, there is no session that is typed exhibition.
You can verify it by searching [time table response](https://deploy-preview-49--droidkaigi-api-dev.netlify.com/2020/timetable/index.json).

ios-base/DroidKaigi 2020/Session/Views/SessionViewController.swift
```diff
@@ -66,7 +69,17 @@ final class SessionViewController: UIViewController {
         let dataSource = SessionViewDataSource()
         let filteredSessions = viewModel.sessions.asObservable()
             .map { sessions -> [Session] in
```
ry-itto: Please use `[weak self]` to prevent circular reference 🙏 
For example, 
```swift
.map { [weak self] sessions -> [Session] in
    guard let self = self else { return [] }
```
And you don't need to fix following `self` 👍 

This is preference. So you don't have to fix this.

clockvoid:
@ry-itto 
Thank you for reviewing.
I fixed it.

ry-itto:
LGTM 😄 
I will merge this 👍

PR #726, opened 2020-02-06
Fetch floor map image from server instead

### LINKED ISSUE #717: Use server side image for floor map(Android)
## Kind (Required)

- Improvement

 ## Overview (Required)

- Probably, we can not prepare image early. So  we should use server-side image
https://api.droidkaigi.jp/images/2020/map.png

 ## Links	

-

kuramu1108:
## Issue
- close #717 

## Overview (Required)
- Use server-side image instead for floor map image


kuramu1108:
I would like to be advised a better placeholder image from the designer(?) maybe

feature/floormap/src/main/res/layout/fragment_floormap.xml
```diff
@@ -17,7 +17,6 @@
             android:layout_height="match_parent"
             android:paddingStart="16dp"
             android:paddingEnd="16dp"
-            android:src="@drawable/ic_floor_map"
```
takahirom: Thanks!
Can we delete ic_floor_map and ic_map_black? Because the location of the room in this image is completely different from the real one, it will be confusing if displayed. 🙇

kuramu1108: definitely ok for the former, but since `ic_map_black` is for the placeholder image, can I get a better one from the designer or any suggestion

kuramu1108: @takahirom actually the commit is deleting the imag usage already, so it is not showing to the user
I suppose you are saying to remove the resource itself as well? 
Since it is still used in
https://github.com/DroidKaigi/conference-app-2020/blob/e10221e0fbb801adf70920a7d3ce5dc7b478cfcb/feature/floormap/src/main/java/io/github/droidkaigi/confsched2020/floormap/ui/FloorMapFragment.kt#L49

which might be removed later on?

takahirom: > definitely ok for the former, but since ic_map_black is for the placeholder image, can I get a better one from the designer or any suggestion

I used it if the image was available, but it seems to be difficult, so I am in this state. 😭 
Nothing is going to be the best I can think of.

> actually the commit is deleting the imag usage already, so it is not showing to the user
I suppose you are saying to remove the resource itself as well?

I want to delete all the images once. The main concern is that this image has been released and the location of the room is different, so participants are confused on the day. 🙇 
If possible, if the URL can be changed in the room, it may be possible to make an image for each room later.

takahirom: New images may be available. Please wait for a while. 🙇

takahirom:
The new map image is available now on Figma!  (CC: @ry-itto )
Can you use it for a placeholder? (Server-side image is still needed)

takahirom:
probably ./gradlew :feature:floormap:ktlint  become error 👀

kuramu1108:
@takahirom  thanks, got time to fix it just now 🙇‍♂️

takahirom:
Thanks! LGTM!!

PR #741, opened 2020-02-08
Fix a bug that the BottomAppBar is not close to the FAB

### LINKED ISSUE #62: Bottom app  bar padding is not exist on Android 5.0

yurihondo:
## Issue
- close #62 

## Overview (Required)
The cause of this issue is that the BottomAppBar cannot calculate its height correctly in API level 21 environment.
So I added process to measure own size when #doOnNextLayout is called.

## Links
- N/A

## Screenshot
| | Before | After
| :--: | :--: | :--:
| Nexus5 Android5.0 |  |


yurihondo:
@takahirom 
Could you please check this PR when you have time?
Thanks.

takahirom:
Thanks!!! Awesome!!!!!

PR #769, opened 2020-02-12
Use new map image

takahirom:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
-

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Sorry, we should use a new image for iOS release!!! 🍎

PR #771, opened 2020-02-13
Fix a bug that the session title is not translated

yurihondo:
## Issue
none

## Overview (Required)
Fixed a bug that the session title was not translated even if set english lang to device.

## Links
none

## Screenshot
| Before | After
| :--: | :--:
|  |


takahirom:
Can we use this?? 👀 
https://github.com/DroidKaigi/conference-app-2020/blob/master/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2020/model/LocaledString.kt#L8

yurihondo:
Oh...Thank you for your advice.
I'll modify it.

PR #773, opened 2020-02-13
[iOS]Change call location of loadMap() 

tomu28:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- `viewDidLoad()` method that must be called in A was called in `init()` , so it was fixed
  - Before the fix,  `Thread 1: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value `  [occurred in the following parts:](https://github.com/DroidKaigi/conference-app-2020/pull/773/files#diff-4bb5ebce785a29b29305cf99ad3c3338R33-R34)

```Swift

let width = imageView.image?.size.width
let height = imageView.image?.size.height
```

## Links
None

## Screenshot
No UI changes.

Before | After
:--: | :--:
 |


ry-itto:
Oh I'm sorry for missing in review ...

tomu28:
You’re very welcome.
Thank you for the quick confirmation!😊

PR #791, opened 2020-02-15
Fix bottom bar wrong position when reattached

FujiKinaga:
## Issue
- N/A

## Overview (Required)
- When open session detail screen and go to floor map by bottom app's menu, bottom app position is shifted.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
LGTM🔥

PR #805, opened 2020-02-16
[iOS] Fix error if build directory is already exists

mkeeda:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Fix error of `mkdir $SRCROOT/build` if build directory is already exists
```
mkdir: /Users/.../conference-app-2020/ios-base/build: File exists
```
## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Looks good

PR #806, opened 2020-02-16
Suppress const modifier lint warning on Dep.kt

chibatching:
## Issue
- close Nothing()

## Overview (Required)
- Suppress `const` modifier lint warning to `Dep.kt`

## Links
- none

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/buildSrc/src/main/java/dependencies/Dep.kt
+++ b/buildSrc/src/main/java/dependencies/Dep.kt
@@ -1,6 +1,6 @@
 package dependencies
 
-@Suppress("unused")
+@Suppress("unused", "MayBeConstant")
 object Dep {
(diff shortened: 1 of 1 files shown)


takahirom:
Because it is simple, I prefer to suppress without const, but is it better to do so? 👀

chibatching:
Yes, it has just a bit performance advantage.
If you prefer to suppress warning, I'd revert the code and change suppress warning.

chibatching:
Hi, I changed the way to remove warnings.
I added suppress rule `MayBeConstant` to `Dep.kt`, and organize commit by rebasing.

takahirom:
Thanks! LGTM 🙏

PR #810, opened 2020-02-24
Use container transform animation

takahirom:
## Overview (Required)
- I tried it

## Links
- https://material.io/develop/android/theming/motion/#material-container-transform

## Screenshot
Before | After
:--: | :--:
before | speaker


feature/session/src/main/res/layout/fragment_session_detail.xml
```diff
@@ -31,19 +31,12 @@
             android:layout_gravity="center"
             />
 
-        <androidx.core.widget.NestedScrollView
```
takahirom: I  think this nestedscrollview is not needed.

